PR #139, opened 2015-05-03
Free up space before install if using Rev A or B BBB

darylc:
Install was failing on pre Rev C BBB's due to the internal MMC being 2Gb rather than 4Gb.

Removes a heap of packages that shouldn't be needed if you are using an old Rev BBB and also cleans up the apt-cache.  Leaves /  around 90% full after installation completes now.


cpinkham:
Thanks.  I'm closing this since I manually applied the remove commands and also refactored some of the other code to try to save space.  The image still could use a lot of TLC to get it slimmed down, but it is small enough to fit on a 2GB SD or eMMC now.  Our FPP v1.5 SD and eMMC images will use a 1.8GB image size.

PR #141, opened 2015-05-10
Updated fppLCD.py to use /home/fpp/media if available

jaredb7:
The current script doesn't work if you use the 'FPP 2.0 install script'
to build FPP install (I have a RPi 2), and this is because the LCD
settings path for self.PLAYLIST_DIRECTORY and self.SETTINGS_FILE is
hardcoded to '/home/pi/media'

I added a little check at init to use '/home/fpp/media' if it exists,
otherwise it will just use the default value of already set in the
script.


jaredb7:
closing due commits being all messed up because I didn't create a branch for this fix.

I'll request another pull with the initial fix

PR #174, opened 2015-10-28
fpp.js refactor and fppjson.php additions

rawcreative:
Rewrite of GetFPPstatus in fpp.js to prevent bottlenecks for slow response times from the pi. Also enables the use of the page visibility API in supported browsers to stop polling when the browser window/tab is in the background/hidden. 

The original GetFPPstatus in both fpp.js and fppxml.php have been left in place for the time being (in case they are used in a plugin or elsewhere).

### DIFF
--- a/www/fppjson.php
+++ b/www/fppjson.php
@@ -250,6 +250,8 @@ function parseStatus($status)
 			'media_filename'    => $status[4],
 			'seconds_elapsed'   => $status[5],
 			'seconds_remaining' => $status[6],
+			'time_elapsed' 		=> getTimeElapsed($status[5])
+			'time_remaining'	=> getTimeElapsed($status[6])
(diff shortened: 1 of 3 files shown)


cpinkham:
Thanks.

I merged this pull request manually all except for the last commit regarding the formatting.  This seemed to have eliminated all the tabs in the file, I'm not sure if that was intentional or not.  I think this is why it had a major conflict with current master when I tried to merge the pull request and since it was cosmetic I decided it could wait until confirming.  The newer code in FPP is almost always using tabs except when in a file that is already using mixed indention styles and in that case I have tried to match.  I'm all for fixing these formatting differences though.  I have a .vimrc file with tabstop=2 to ease the pain of editting some of these files.

Issue #184, opened 2015-12-28
Enhancement to allow external control over GPIO


rudybrian:
If FPP and external applications (using the WiringPi libraries) attempt to read/write GPIO pins they will likely clobber each other, even if they are not attempting to manipulate the same GPIOs due to the way WiringPi is implemented (See [here](http://falconchristmas.com/forum/index.php/topic,3278.0.html) for example). Since many users need to use GPIOs outside of FPP, this enhancement provides a means of sharing the GPIOs between FPP and external applications by using FPP as the arbiter, and providing commands to set the mode and read/write/update GPIO ports from external applications while fppd is running.

rudybrian:
Initial tests in v1.7 (after adding 418895a80b66b0953d698c2ec8963f7bde17042e, 8a1bc626596e56b70f5d14da61e29ac58d9056f9 and a5e221a670ea3fb17804ed3b8292fa4f64d07531) look good, but more testing is needed for the PiFace.

rudybrian:
I finally had a chance to test this out with the PiFace after pulling my display down last weekend and everything looks good.

rudybrian:
This enhancement has been committed as of #186

PR #186, opened 2015-12-30
Add support for shared GPIO control (#184)

rudybrian:
Things appear to be working on my end with the onboard GPIO, but could use some additional testing with SPI-based GPIO expanders.


rudybrian:
This code has now been tested with the PiFace and no problems were observed.

cpinkham:
I haven't tested it, but it looks good from what I can see.  Thanks for working on this.

PR #188, opened 2016-01-04
Update uploadfile.php to be Apache Proxy "friendly"

bigwavedave42:
The "/" at the beginning of the URLs were not relative paths, breaking things when in an a proxy configuration. ex: https://www.example.com/falcon/

Using a proxy configuration allows me to add SSL, ACLs, and authentication when accessing the UI remotely.

On the File Manager page, the Upload Files section would not show the "Select Files" button.

NOTE... this is my first "pull request" and am open to feedback if there is a better protocol.


cpinkham:
Thanks, pull request was fine.

Issue #221, opened 2016-04-11
exfat-utils not installing on upgrade from 1.5 to 1.8


bigwavedave42:
I did a fresh install of v1.5 and did an immediate upgrade to v1.8.  On console after the upgrade I noticed:
```
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following extra packages will be installed:
  exfat-utils
The following NEW packages will be installed:
  exfat-fuse exfat-utils
0 upgraded, 2 newly installed, 0 to remove and 0 not upgraded.
Need to get 109 kB of archives.
After this operation, 286 kB of additional disk space will be used.
Err http://mirrordirector.raspbian.org/raspbian/ wheezy/main exfat-fuse armhf 0.9.7-2
  404  Not Found
Err http://mirrordirector.raspbian.org/raspbian/ wheezy/main exfat-utils armhf 0.9.7-2
  404  Not Found
Failed to fetch http://mirrordirector.raspbian.org/raspbian/pool/main/f/fuse-exfat/exfat-fuse_0.9.7-2_armhf.deb  404  Not Found
Failed to fetch http://mirrordirector.raspbian.org/raspbian/pool/main/e/exfat-utils/exfat-utils_0.9.7-2_armhf.deb  404  Not Found
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```
I realized it was likely stale info on the image, so I did a `sudo apt-get update` followed by a `sudo apt-get install exfat-fuse`.

It was able to successfully install then.  

You'll notice the packages have been updated, as they are now pulling:
```
Get:1 http://mirrordirector.raspbian.org/raspbian/ wheezy/main exfat-fuse armhf 0.9.7-2+deb7u1 [27.1 kB]
Get:2 http://mirrordirector.raspbian.org/raspbian/ wheezy/main exfat-utils armhf 0.9.7-2+deb7u1 [83.6 kB]
```

I'm guessing "anytime" an apt-get is done in the script, a `sudo apt-get update` should happen before it.... or at least with some freshness considerations.  This would resolve these problems in the future.

PR #223, opened 2016-04-11
Refresh apt cache before install

bigwavedave42:
apt might need to be updated before installing new packages, as seen in issue #221


bigwavedave42:
It looks like this was effectively done in commit a212419c09af2ab68fb6baae964bea77bcf64390

bigwavedave42:
This has been resolved by commit a212419c09af2ab68fb6baae964bea77bcf64390

PR #249, opened 2016-11-01
70-persistent-net.rules not found throws sed error

randyr505:
I updated the script to check if file exists prior to running sed against it. However, I am assuming this file should exist if it is being edited and is on a fresh FPP image.


cpinkham:
For some reason I had to manually apply the fix to the fix, but this change is now in master and master-v1.x.

PR #254, opened 2016-12-28
Support for ETH1/WLAN1 and FPP Access Points

jcochran:


jcochran:
I've added commit to the pull request to add a functional Access Point mode to the FPP. Its limited in scope ideally to support people using tablets/phones to access the show network, master to remote FPP, or master to F16V3. It's not intended to support sending E1.31 data and may not support all WiFi adapters out there. I know for sure it does support the EdiMax adapters as they require a custom driver and a custom version of hostapd and I'm not going down that rabbit hole again right now. More details on how to config the access point are in the commit notes. I would only look at commits ed35354 which adds additional interface support and 004dcb8 for Access Point. The stuff in the middle was for network builder which gets backed out with commit ed8f6ec.

cpinkham:
Do you have time to test and update this to see if it works with v1.9 and master-v1.x?  If so, I can look at getting it merged into master-v1.x and a potential v1.10 soft release before the end of the year.

jcochran:
I don't have any of my stuff out at this time not even my development assets, too overwhelmed with other adventures currently. I was going to start pulling things out after the New Year I need to work on so I'll be able to take a look at things then.

jcochran:
Related to this, I have an opportunity over the summer to try to scale up a Raspberry Pi with HostAPD for my shooting gallery project and found the it maxed out with 8 client connections and then blocked any further connections, so I to punt and pull out a network using UniFi devices since I had to scale to 50+ connected devices even though it was very low bandwidth. I believe the limitation is with the wifi adapters/drivers themselves, so this functionality would be limited in scope to only a handful of connected devices which kinda fits into how I was coming at with this latest version I had here in the pull request.

cpinkham:
Closing this pull request.  This was for v1.x and FPP v2.x has major changes including using connman for interface management and support for tethering via WiFi if no other interfaces are up on the Pi/BBB.  Some of these ideas may still hold merit for inclusion in v2.x, but will likely need reworking and definitely need a different pull request against FPP's master code branch.

PR #270, opened 2017-02-09
Missing Email System Fix

### LINKED ISSUE #267: Email Support is Broken
Email support is broken in the v1.8 image/baseline.

- The 1.8 image is missing the exim4 packages, as well as the command mail tools like mail/mailx.
- There is no trace of exim4/mail in the 1.8 install script to either install these packages or configure them post-install (e.g. setup the smtp.gmail.com smarthost for relay).
- I can find no other place which references exim, mail, or gmail outside of the GUI files so I don't think the missing files are being added/configured anywhere else.
- There is a bug on line 392 column 42 in common.php that prevents the FPP created aliases file from being copied to /etc (e.g. there's an extra period).
- The created aliases file still uses the original pi user and not the new fpp user (line 390).
- The chfn command (line 386) still uses the original pi user.

jaredb7:
Fixes an issue with exim4 is missing from the FPP system, the upgrade script installs and configs exim4 for gmail.
Small change to the common.php SaveEmailConfig function targetting the fpp user and not the pi user

closes #267


cpinkham:
I applied this pull request manually to master and master-v1.x.  Closing request.

Thanks, and sorry for the delay.

PR #271, opened 2017-03-03
F4b dmx

dkulp:
This is a bunch of updates and optimizations of the F16B/F4B backend code to get DMX and all the pixel outputs (where possible) working together.    I have tested with F4B and F16B and DMX now works cleanly with all pixel outputs enabled.  It shoudl work for up to 40 outputs enabled, but the current GUI doesn't allow that.

There is likely some changes needed to the PixelNet code related to the setup registers like was done for the DMX, but I have no PixelNet anything to be able to test it.


dkulp:
Just closing as all the changes are on the stretch branch along with a ton of other things

PR #285, opened 2017-10-17
Add colorlight and linsn outputs to the master-v1.x branch

rudybrian:
Since a bunch of folks have been asking about Linsn and Colorlight support in FPP, I thought I would have a go at backporting the a few of the new channel output types @cpinkham added to the master branch into the master-v1.x branch so we could start taking advantage of them before the release of 2.x.

Things seem to be working, but I need some folks with known working configurations to test.


rudybrian:
I managed to mess up my master-v1.x tracking branch somehow, and it had become a bit stale. The latest commit pulled in all the more recent changes from master-v1.x and now the unresolved conflict warnings are gone.

PR #288, opened 2017-11-07
Fix BBB RTC not registering correctly issue - bbb-stretch-v1.x

jaredb7:
added a sleep delay to piRTC to let thing settle down before using hwclock

changed i2c buses for setupBBBRTC the method in functions.
Added crude if statement to check if result from previous command is non 0. tested by making it assign i2c-1 first, it fails and does i2c-2. Reversed the order to avoid false positives in the startup log

It seems that in the latest image, and for whatever reason the i2c bus
numbering has switched around. this could be by design or how things are
designed / processed during boot.
Made the script try to configure the i2c-2 bus first then try the i2c-1 bus

In the BBB 1.5 image I did have a RTC working without issue.. I did google issue about the i2c buses changing (1 to 2) found 2 examples but no explanation. I can only assume that this is down to the OS and how it registers/sets up the i2c interfaces

### DIFF
--- a/scripts/functions
+++ b/scripts/functions
@@ -911,11 +911,11 @@ setupBBBRTC () {
 			# BBB has onboard clock as rtc0 so we need to use rtc1
 			echo ds1307 0x68 > /sys/class/i2c-adapter/i2c-2/new_device
 			
-			# i2c busss have swapped around or even by design since the last image
-			# i2c-1 was where the RTC before, nothing can be found on i2c-1 now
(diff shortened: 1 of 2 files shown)


jeffla:
Interesting timing! I ran across this problem last night! Thanks for this.

dkulp:
Can you verify that the latest code on the branch works?

jaredb7:
Hi Dan, the latest code on the branch works. I made this fix about the same time you committed yours.
Don't need this pull request anymore 👍

PR #291, opened 2017-11-27
Allow Ethernet configurations without GW specified

jaredb7:
Allow Ethernet configurations without GW specified, if no GW is
specified with the old code then there is a trailing forward slash and
seems connman freaks out about it. Interface won't get configured.

Have tested going from DHCP. to static with GW and then to static without GW, and DHCP to static with no GW. Seems to work as intended now.


cpinkham:
Closing since this was manually merged in 4d15392.

PR #292, opened 2017-11-30
[Security] [UI] Implement basic authentication via PHP

nkaminski:
UI authentication does not work since Apache to nginx migration. This changeset implements HTTP Basic authentication within the PHP code such that there is no longer any dependence on the HTTP server software itself for authentication. Furthermore, this could easily be modified to allow more granular access control, as well as to make certain pages always publicly accessible if desired.

### DIFF
--- a/www/about.php
+++ b/www/about.php
@@ -1,5 +1,6 @@
 <?php
 require_once('config.php');
+require_once('auth.php');
 
 $a = session_id();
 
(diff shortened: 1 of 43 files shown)


cpinkham:
Closing this pull request since we decided to stay with Apache for FPP v2.x.  If we ever go back towards nginx, we can revisit this.

nkaminski:
OK, sounds good.

Issue #283, opened 2017-10-06
Multisync.php stops responding on RasPi 3 


JonB256:
Running v1.9-64-g01a089f on all three. They are all connected via wlan0 using a Static IP. There is no CAT5 connection. 

After "a period of time" they stop being visible to my FPP Master (which is a Pi 2B), They are fully accessible by their IP addresses but will not even display Multisync data on their own pages. 

I have found that going to their Network page and Refreshing the network always brings them back to full function immediately (but not permanently). 

This has happened repeatedly but only on RasPi 3 models. This is being discussed on the Falcon Forum also.

patdelaney:
I have a similar issue on a Pi B running Jessie and master v1.x branch. When I go to multisynch tab it eventually times out and can't get the status or a unit. I have a mixture of B, B+ V2, and V3. All are currently in standalone mode. I have seen it fail to discover v2 and v3. I can open the multisync page on another model B running Wheezy 1.9 and the page dispays fine, no issues.
![multi-synch_error](https://user-images.githubusercontent.com/8147459/32144829-9d74e404-bc8c-11e7-8ea8-d4c1312eb07d.jpg)

jaredb7:
@patdelaney would you happen to have the log files from this install at all? Mainly the apache2-error.log from the /logs directory on the USB stick 
I just want to see what's causing the Get Status Failed error, I run some of my FPP's with a GUI password and you'll get this error if that's the case, but want to see what the issue is for others.

possibly related to #356

patdelaney:
I enclosed the nginx log  this is a 2.x system 

 

Pat

jaredb7:
Hi Pat, log didn't come through. But found the cause for multisync timeout.
I've posted a fix for the issue in 2.x, will also add the fix into 1.x also.

cpinkham:
Believe this is fixed in latest FPP v3.5.x

PR #360, opened 2018-03-13
Multisync Page hang / Stops responding fix

### LINKED ISSUE #356: FPP Multisync page
Go to multisync page, click on a link to go to another FPP, do some work on the other fpp screen and then use browser arrorw to go back to the original FPP instance. The web UI is locked up.

Even if you try to open a new browser. The Webui is "stuck"  I have to restart nginx service

jaredb7:
Issue is to do with PHP sessions and session locking. nginx runs scripts super quick and we end up with read locks on the PHP session. This is worse when you have a lot of FPP devices that need polling, we end up with 6+ get getFPPStatus requests all firing off at once

Fix is to close the session before we start any FPPStatus work (we don't need it anyway) and at the beginning of returnJSON.
Closing the session in getFPPStatus fixes the issue, attempting to mitigate edge cases with closing in returnJSON also.

Added in minor handling of errors with file_get_contents
Added reason for a failed read when the remote device has a WebGUI password via a 'unknown' status
image

Will backport this into v1.x as it was happening there, rarely but it did happen - also session related under apache

closes #356 
related in part #283

### DIFF
--- a/www/fppjson.php
+++ b/www/fppjson.php
@@ -78,7 +78,10 @@ return;
 /////////////////////////////////////////////////////////////////////////////
 
 function returnJSON($arr) {
-	header( "Content-Type: application/json");
+	//preemptively close the session
+    session_write_close();
+
+    header( "Content-Type: application/json");
 
 	echo json_encode($arr);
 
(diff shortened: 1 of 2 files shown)


dkulp:
This has been merged for a while

PR #363, opened 2018-03-31
 [Scripts] Modify git_pull script to correct perms 

jaredb7:
Modify git_pull script to correct permissions on the /opt/fpp/upgrade
directory, newly created folders are missing the execute bit and that
upgrade file doesn't execute.

Added in a line to chmod that directory & subdirectories before running
upgrade_config


cpinkham:
It looks like this was only an issue with the '30' directory and I didn't catch it when I merged the pull request.  The 30/upgrade.sh script now has the proper execute permissions in the git repository.  I updated the FPP_Install.sh script so the php-zip package is included in the base image so this 30/upgrade.sh script won't be necessary in the future starting with the next Pi SD image build.

cpinkham:
If it becomes an issue, we could issue a 31 upgrade fix.

We probably should also update the upgrade_config to only update the version number in /etc/fpp if the upgrade.sh script completes successfully.  If you want to test that fix to upgrade_config and submit a pull request, it would be appreciated.

PR #418, opened 2018-09-19
[Enhancements] Multisync Page (expert/advanced view) + Host Description

### LINKED ISSUE #392: v2.x Enhancement Add a Description field in settings or network tabs
While I am probably not the norm when it comes to users. I would like the ability to add a description somewhere in the UI that says what the Pi/BBB is used for. It would help people that have large shows or don't name their FPPs with their purpose.

Enhancement for v2,x+1  Return the new  description field in the Multisync page when you are looking at available systems.  At some point I would like a "Dashboard" of all the running systems. that I can look at.

I could write up what I am looking for and maybe get Ben to do it if we move forward with this enhancement.

jaredb7:
Adding a 'expert' view to the multisync page. Which reveals a few more columns showing stats about the device.

- Version - what base version the device is (eg 2.0)
- Git branch - what branch the device is on (eg. v2.0 or master.. etc)
- Git versions - remote i.e on github, and local i.e on the pi/BBB, same as what you see on the About page
--- Additionally if there is a version difference a "Update Available" link will show, and will link to the relevant system's about page
- Auto update checkbox status
- Utilisation with CPU, Memory & Uptime

Network Settings
Added ability to set a host description, can be used to label the device.
This is shown on the Multisync page for all devices where set

Updated the getSysInfo JSON API endpoint to return extra stats like git versions, and utilization stats.
New JSON API endpoint to get the hostname and host description

Moved a few things like commands to get git branch and git versions, memory usage, cpu usage into common.php

Closes #392

image


patdelaney:
Awesome,,, thanks

dkulp:
Can you please make sure you are basing pull requests off of master?   A "git rebase origin/master" type thing should do it.

dkulp:
I've merged this onto master.   I also added the variant ("PocketBeage", "Pi Model 3", etc...)  under the platform in expert mode as that's available in the getSysInfo.

One concern I DO have that will require keeping an eye on:   previously, every page refresh involved a single call to each remove to get the status.   Now, it's two (status and host description) with a possibility of a third (expert info).  Not sure what impact that may have on the running remotes.   Kind of wondering if the description (and variant) could/should be added to the Avahi service record and parsed to avoid the extra call to the remotes.

Issue #506, opened 2019-03-16
Playlist:NextItem(), Playlist:PrevItem() calls Clean() when it shouldn't


ghormann:
While testing the MQTT changes, I noticed that PlayList:NextItem() calls StopNow() which calls calls SetIdle() which calls Cleanup().   Cleanup will then delete all the PlaylistEntryBase records.   When control returns back to NextItem(), it tries to call Start().  Start() fails because there are no PlaylistEntryBase .   As a result, we end up in the Idle state.

Can't the call to Cleanup() in SetIdle() just be removed?  It is already called from Load(Json:Value) to prevent the memory leak.

ghormann:
I did some further testing of and don't think Idle() should call Cleanup().   Removing caused NextItem() and PreviousItem() to work as expected.  I didn't find any other problems when testing via GUI or MQTT.  Thus, raising pull request #507

dkulp:
This has been fixed for a wile.

PR #507, opened 2019-03-16
ISSUE506: Don't delete PlayListEntries on call to Idle()

ghormann:
I did some further testing of #506 and don't think Idle() should call Cleanup().   Removing caused NextItem() and PreviousItem() to work as expected.  I didn't find any other problems when testing via GUI or MQTT.

### DIFF
--- a/src/playlist/Playlist.cpp
+++ b/src/playlist/Playlist.cpp
@@ -765,7 +765,8 @@ void Playlist::SetIdle(void)
 	m_sectionPosition = 0;
 	m_repeat = 0;
 
-	Cleanup();
+	// Remoted per issue #506
+	//Cleanup();
 
 	if (mqtt)
 	{


cpinkham:
Sorry for not commenting on this yet.  The problem with this code revert is that now playlist entries are not deleted until the next playlist is loaded.  That was the bug that was fixed in 3f930e11c by adding the Cleanup() call.   The real issue is that ::NextItem() and ::PrevItem() should not be calling StopNow() which sets the idle state.  We shouldn't stop a whole playlist and restart it in order to go to the next item in the playlist.  The next/prev code should be stopping the current item playing (while leaving the playlist itself running) and adjusting the pointer to point to the next/prev item then starting that item playing.  Stopping and restarting the whole playlist in next/prev can have other ramifications which are undesirable.  I will take a look at fixing the next/prev calls.

ghormann:
I understand your second point about why we shouldn't stop the whole playlist, but I don't understand the first point. Why should the playlist entries be deleted when stopped.  Can you explain?   

The only difference I noticed was with the UI. Personally, I like how the UI doesn't delete the entries on stop now with this change.  I just assumed that deleting them was a bug...  When you select a new one, the new ones display before you play. 

Thanks for helping to educate me....

cpinkham:
Currently, we have one Playlist singleton, but that will change at some point in the future.  The Playlist rewrite was done with the intention of allowing multiple playlists to be playing at the same time.  I also have an old patch which allows playing multiple sequences at the same time to support this.  Currently, since we have the single Playlist instance, if it isn't playing, its data should be cleaned up.  Without this Cleanup() call, when a playlist stops, it holds the playlist itself open in memory with all of the PlaylistEntry* object instances.

As part of the recent playlist work, I had updated the UI to pull in the currently playing playlist data from the Player itself, so we picked up any dynamic updates and also because long-term, you will be able to insert items into a running playlist.  The fix also allows the UI to automatically update if you changed a playlist on-disk and the player picked up the changes at the top of the next loop.

It sounds like I need to fix the UI so that when the playlist stops playing, the player reloads the playlist from disk instead of picking up the now-blank playlist from the player.  This was not an issue previously when the UI never updated the playlist except when you reloaded the page or selected a playlist from the dropdown select.

ghormann:
Thanks Chris.  I think I understand now.   To me, freeing up the memory wasn't that important since we'll probably play the playlist again, but given this run on small memory devices maybe that isn't the best thought process.

Sounds like what we need is a Pause() method or maybe just another optional parameter to stop on if setIdle() should be called.   This could then be used with Prev/Next functions.   

If you agree, i can work on a PR.

PR #555, opened 2019-09-19
Actually add debug symbols for "debug" target

ghormann:
Debug  target should actually produce debug symbols. :-)


dkulp:
This is not needed on master as the -g flag is already added there with "make debug".

PR #566, opened 2019-10-10
Updated Co-Other for new output types

nick-diel:
I have started on some projects for adding additional output types and saw an opportunity with co-other.php to hopefully make it easier and quicker to add new output types to the web frontend.  
My goals for these changes were:
1.  Not to fundamentally change how the table was being built
2.  Leave the code for the existing output types as is, as well as the actual code for building and saving the tables.
3.  Using Javascript's ES6 class functionality, have a quick method (reusing code) to create a new output types.
4.  Eliminate the need to update the create, populate, save, etc. functions in co-other everytime one adds a new output type 

In addition there are a few other small changes in my branch
1.  Backend and frontend code for a generic SPI output
2.  Using the new method added a frontend code for MSCP23017
3.  Made a small update in the backend code for MSCP23017 to remove the hard coded I2C address
4.  I moved the SPI-WS2801 frontend code to the new method as it was my test case.

Let me know if these things should be in their own pull request.

### DIFF
--- a/src/channeloutput/generic_spi.cpp
+++ b/src/channeloutput/generic_spi.cpp
@@ -1,0 +1,195 @@
+/*
+ *   Generic SPI handler for Falcon Player (FPP)
+ *
+ *   Copyright (C) 2013-2018 the Falcon Player Developers
+ *      Initial development by:
+ *      - David Pitts (dpitts)
+ *      - Tony Mace (MyKroFt)
+ *      - Mathew Mrosko (Materdaddy)
+ *      - Chris Pinkham (CaptainMurdoch)
+ *      For additional credits and developers, see credits.php.
+ *
+ *   The Falcon Player (FPP) is free software; you can redistribute it
+ *   and/or modify it under the terms of the GNU General Public License
+ *   as published by the Free Software Foundation; either version 2 of
+ *   the License, or (at your option) any later version.
+ *
+ *   This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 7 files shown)


src/channeloutput/generic_spi.cpp
```diff
@@ -0,0 +1,195 @@
...
+/*
+ *
+ */
+
+int GenericSPIOutput::Init(Json::Value config) {
+    char configStr[2048];
+    ConvertToCSV(config, configStr);
+    return Init(configStr);
+}
+
```
dkulp: Just a note about this:    ideally for "new" code, you would use the JSON directly for configuring the output and not convert to CSV.   The CSV stuff is legacy and the ConvertToCSV method is there to make migrating the outputs from 1.x quicker/easier.   Newer outputs should just use the JSON.     (and any help updating old code would be appreciated)

dkulp:
Squashed and merged onto master....  Please base future pull requests on master.   Major thanks for the contribution.  The web page cleanup is awesome.   So much easier to understand.

patdelaney:
I'm late on the thread, does the documentation need to be updated with new
screenshots?

Pat

dkulp:
No...  There is a new "Generic SPI" option for output type on the "Other" tab that could be documented, but no appearance changes other than that.

src/channeloutput/generic_spi.cpp
nick-diel: Got it.  I'll update some of these modules to use JSON.

PR #628, opened 2019-12-23
Bypass ui password if request is from local gateway

fearfair:
Only require ui login if attempting to access from outside the local lan.  Since all traffic via port forward will show up as the router's ip, only require login when request originates from the router ip.


dkulp:
This needs to be rebased on master.... Way too much going on in this PR right now.  :(

dkulp:
This pull request doesn't contain any actual changes related to the description.

PR #640, opened 2020-01-16
enh: improve V2 FSEQ header writing

Cryptkeeper:
This PR simplifies the V2 FSEQ header writing process.
- Moves magic numbers into const fields to reduce duplication and potential mismatches. (I'm not familiar with C++ idioms, should these be preprocessor definitions instead?)
- Builds header as a single buffer. This avoids allocating additional buffers and write calls for things like writing sparse ranges, compression blocks, variable headers and alignment bytes.
- Shuffle field assignments to be in descending order by index for code readability.
- Improve write validation to note alignment mismatches or overflows.
- Fix incorrect documentation and add new comments & references.
- Prevent defaulting behavior from matching invalid major version values to major version 2. (Since `FSEQFile.cpp`/`FSEQFile.h` are used by xLights, I've cross checked their usage and they already properly validate the version parameter input, so this is a non-issue for their implementation.)
- Added missing newline breaks to fsequtils to correct output errors with bash & zsh.
- Fixes fsequtils segfaulting when outputFilename is null (which occurs whenever you simply want to print verbose information about a file without outputting to a new file).
- Fixes uncompressed V2FSEQ files from printing a corrupted file warning when `m_sparseRanges` is empty.

### DIFF
--- a/src/fseq/FSEQFile.cpp
+++ b/src/fseq/FSEQFile.cpp
@@ -1392,21 +1392,24 @@ void V2FSEQFile::writeHeader() {
 
     uint8_t maxBlocks = m_handler->computeMaxBlocks() & 0xFF;
 
-    // Compute headerSize to include all data except channel data
-    // This allows the use of a single buffer, and single write to disk
+    // Compute headerSize to include the header, compression blocks and sparse ranges
     int headerSize = V2FSEQ_HEADER_SIZE;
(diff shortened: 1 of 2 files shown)


Cryptkeeper:
I've left this PR as a draft until I have time to fully smoke test it :)

Cryptkeeper:
I've opened it for review. There's still more I plan to do, but I want to avoid an overly large PR especially since I'm currently shooting in the dark in terms of idioms and code style.

PR #655, opened 2020-01-20
enh: cleanup V1/ESEQ file reading

Cryptkeeper:
- Adapts my previous V2FSEQ#writeHeader simplification & optimization PR (https://github.com/FalconChristmas/fpp/pull/640) to V1FSEQ#writeHeader
- Moves ESEQ related magic numbers to named constants
- Reorganized V2FSEQFile's buffer constructor to read by index in ascending order
- Added read validation to prevent index overflows
- Fixed outdated comments and added additional comments explaining preexisting behavior
- When peeking into a FSEQ file, it reduces the initial read to 8 bytes from 48. 8 bytes ensures it only loads the necessary fields (format identifier, version fields and channel data offset) instead of the full header plus garbage data. As it did previously, the code still does an additional read for the full header once it has validated the initial 8 byte peek. Reading up to 48 instead of simply 8 isn't useful since it re-seeks back to index 0 and re-reads, and there is still fadvise calls for the initial 1kb of the file.
- Fixes a DumpHeader error where it was using a previous buffer but an updated bytesRead value.
- Tightens writePos validation to match the exact value instead of a less than comparison.

### DIFF
--- a/src/fseq/FSEQFile.cpp
+++ b/src/fseq/FSEQFile.cpp
@@ -552,9 +552,9 @@ void V1FSEQFile::writeHeader() {
         writePos += len;
     }
 
-    // Validate final write position does not exceed channel data offset
-    if (writePos > m_seqChanDataOffset) {
-        LogErr(VB_SEQUENCE, "Final write position (%d) exceeds channel data offset (%d)! This means the header size failed to compute an accurate buffer size.", writePos, m_seqChanDataOffset);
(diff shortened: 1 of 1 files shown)


Cryptkeeper:
Bumping so this doesn't become stale. If there's a more preferred way for me to be doing these PRs let me know. @dkulp

Cryptkeeper:
Bumping since this has passed a month since being opened. @dkulp

PR #677, opened 2020-02-22
Removed "divisible by 16" check

computergeek1507:
I manually edited the "co-other.json" config file with different resolutions and everything worked as expected. I don't see why this is needed.

### DIFF
--- a/www/co-other.php
+++ b/www/co-other.php
@@ -75,12 +75,6 @@ var FBDevices = new Array();
 
 function VirtualMatrixLayoutChanged(item) {
 	var val = parseInt($(item).val());
-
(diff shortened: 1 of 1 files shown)


cpinkham:
What model Pi was this tested on?  I think that this code was added originally because the earlier model(s) couldn’t create a framebuffer that had a width or height that was not a multiple of 16.

Chris

computergeek1507:
Raspberry Pi 3B non plus version

cpinkham:
Tested working on a Pi 2B as well, so either issue was resolved by newer firmware/OS or it was only ever an issue on the original model B.

PR #760, opened 2020-07-23
Added ZCPP Output, Channels data only, no configuration data

computergeek1507:


src/channeloutput/UDPOutput.cpp
```diff
@@ -254,6 +255,10 @@ int UDPOutput::Init(Json::Value config) {
                 //DDP types
                 outputs.push_back(new DDPOutputData(s));
                 break;
+            case 6:
```
TechplexEngineer, line 258: I wonder if there is a reason the other output types have two "types" Maybe ZCPP should too?

MarkA75160: My guess is this is why there are two type constants for DDP...
image

Issue #787, opened 2020-08-28
Uncommon sequence Step Time (like 17ms) cause audio and lights to not stay in sync - Fixing commit included


ShadowLight8:
I'm working with a custom controller and am attempting to play sequences at ~60 frames per second or a 17ms Step Time. When the Refresh Rate is calculated by "1000 / Step Time", if the result is not an integer the audio and lights don't stay in sync. Near the end of the audio, the lights continue on for a few more seconds before completing normally.

I changed m_seqRefreshRate and associated usage from an int to a float. Initial testing on my display looks good and things stay in sync.

https://github.com/ShadowLight8/fpp/commit/b0c79c7e7a71b00435844663d2d9104257e45da8

I'm not sure if there is any additional impact from this change, but so far, so good.

ShadowLight8:
@ghormann @dkulp Thanks for merging this in. Closing the issue.

PR #788, opened 2020-08-28
Convert Step Time From Integer to Float

ghormann:
This is to address #787


ghormann:
Change seems reasonable, but prefer @dkulp  review as this is not an area of the code I normally work with.

dkulp:
I merged this as it's definitely an improvement over what's there.   I'd like to actually get rid of the RefreshRate variable entirely and just use the seqStepTime directly.   There really isn't a need to know the FPS for anything, the millisecond step time is really the important thing since that is recorded in the sequence and everything else should be derived from that as needed.

Issue #768, opened 2020-08-01
Increase count in LED Panel Layout "Panel Number" Dropdown...


idleup:
Channel Outputs -> LED Panels.

Dropdown for each panel in the LED Panel Layout Section only goes to "P-16". I currently have 24 panels per output using 32 x 32 P5's which with ColorLight cards are considered on the same Physical output even though they are on 3 different receiver cards. We need that drop-down to go higher like to "P-40" to accommodate panels up to 20-foot across using 32 x 32 panels.

darylc:
I would also like to go more than 16 panels wide for colorlight

darylc:
@ghormann Can we close this one please?

ghormann:
Increased from 16->24 a while ago closing.

PR #845, opened 2020-10-27
Increase panel width limit from 16 to 24 for Colorlight

darylc:
Change has been tested on 18 panel wide Colorlight matrix, 24 is an arbitrary number, much like 16 was.

### DIFF
--- a/src/channeloutput/PanelMatrix.h
+++ b/src/channeloutput/PanelMatrix.h
@@ -30,7 +30,7 @@
 #include "ColorOrder.h"
 
 #define MAX_MATRIX_OUTPUTS    12
-#define MAX_PANELS_PER_OUTPUT 16
+#define MAX_PANELS_PER_OUTPUT 24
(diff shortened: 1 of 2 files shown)


ghormann:
I don't have a Color lights card to test, but this seems very safe.  Merging into master, it can be easily reversed.

darylc:
Thanks @ghormann - I also just saw there is a request to increase it to allow even more panels in https://github.com/FalconChristmas/fpp/issues/768 - not sure how you feel about that?

ghormann:
Honestly, I have no way to test it.    I can't see the increase being harmful to those that don't need so many.  From what I can see it will slightly increase memory usage for everyone to store LEDPanel because of the array in PanelMatrix.h but this is trivial. The pixel memory is dynamically allocated.  

I say go for it.

patdelaney:
I would think the requester should check it.  :-)

Issue #846, opened 2020-10-28
Add  so that we can turn CEC enabled TV's on and off with a script like the projector Plugin


patdelaney:
**Is your feature request related to a problem? Please describe.**

Add the ability to use a script to turn on CEC enable TV's via a script. We need to add the following packages
sudo apt update
sudo apt install cec-utils

Here are 2 scripts we should add to the script repo
on.sh
#!/bin/bash
echo "on 0" | cec-client -s

off.sh
#!/bin/bash
echo "standby 0" | cec-client -s

patdelaney:
I need to document the scripts

darylc:
created pull request to add package to future SD card image builds https://github.com/FalconChristmas/fpp/pull/847

dkulp:
Honestly, to me, this is an ideal candidate to be a new FPP Plugin that uses the ability for plugins to create FPP Commands.    The "install" script for the plugin could install the necessary dependencies.   The commands would then be "TV On"  and "TV Off" (or whatever).   The users using it would have direct FPP commands to use and not need to fuss with running scripts.

patdelaney:
I agree on this one, maybe I'll take a stab at it.  If it's in the pull
request please remove it

MrPaulAR:
I had already written a script but here is a pull request that I created too.  Feel free to close it if this is the wrong answer.  Patrick's scripts above are from me via zoom so I just took the time to add it before looking for enhancement requests.

https://github.com/FalconChristmas/fpp/pull/861

darylc:
Pending someone to write a plugin if anyone desires.

PR #847, opened 2020-10-28
Add user requested packages to future SD image. cec-utils for cec-cli…

darylc:
Add user requested packages for future image release
cec-utils for cec-client - requested in #846
snmp for snmpget/snmpset - requested by me


patdelaney:
I think we might want to look at creating these installs like we do with
Plugins. At least mine is a limited use case and there is no need to
balloon the size of the disk if it's really not needed.

Pat

ghormann:
I'm closing this pull request without merging. As discussed here and in #846  this can and should be done by the plugin installer script.

PR #861, opened 2020-11-06
New Script to Power On/Off a Supported HDMI Device via CEC

MrPaulAR:
Adding two scripts that will power on and off a HDMI connected device if it supports CEC.  This is the same protocol that Chromecasts, FireTV, Receivers, and many other components use.


ghormann:
in #846  it was already discussed that this should be part of a plugin vs base install. 

I'll let @dkulp  Make the final call if to approve or reject the PR.

Jchuchla:
I'll upvote this PR.  I literally just wrote two scripts almost identical to these for my setup.
I had to add the argument for the port at the end before it worked though on mine.  

echo 'standby 0' | cec-client -s -d 1 RPI

MrPaulAR:
Been nearly a year and there are now conflicts so I'm closing this.

PR #917, opened 2020-12-10
Enhanced LOR Support

davidcorrigan714:
Some friends asked me to program a show with a bunch of pixels, and they had all LOR Pixie gear - some new this year some from a show they did last year. I wanted to use xLights and an FPP so I figured out the Enhanced LOR protocol and added the code for it. Tested against the Pixie 4, 8 & PixieCon16 and it's been working well.

I have some notes that would probably be good to go in the manual if someone wants to point me to where that's at to contribute to. The gist is that we used the stock LOR USB adapter for initial testing and used 2 FTDI USB-COM485-PLUS4 adapters for the show to do 6 serial streams for a total of 2600 pixels. Most of the LOR Pixie devices worked up to 1 Mbps though one of our Pixie 4 wasn't cooperating at that speed for some strange reason so we just ran those at 500 kbps and 1 Mbps on a PixieCon 16 since it had a few more pixels on it.

### DIFF
--- a/src/channeloutput/LOREnhanced.cpp
+++ b/src/channeloutput/LOREnhanced.cpp
@@ -1,0 +1,312 @@
+/*
+ *   Light-O-Rama (LOR) channel output driver for Falcon Player (FPP)
+ *
+ *   Copyright (C) 2013-2018 the Falcon Player Developers
+ *      Initial development by:
+ *      - David Pitts (dpitts)
+ *      - Tony Mace (MyKroFt)
+ *      - Mathew Mrosko (Materdaddy)
+ *      - Chris Pinkham (CaptainMurdoch)
+ *      For additional credits and developers, see credits.php.
+ *
+ *   The Falcon Player (FPP) is free software; you can redistribute it
+ *   and/or modify it under the terms of the GNU General Public License
+ *   as published by the Free Software Foundation; either version 2 of
+ *   the License, or (at your option) any later version.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "fpp-pch.h"
+
+#include <fcntl.h>
+#include <termios.h>
+
+#include "serialutil.h"
+
+#include "LOREnhanced.h"
+
+#define LORE_MAX_PIXELS_PER_UNIT     170
+#define LORE_MAX_CHANNELS_PER_UNIT   LORE_MAX_PIXELS_PER_UNIT * 3
+#define LORE_UNIT_MAX_PACKET         518 // 2 for Unit Address & Enhanced Command header, 510 channels, 4 addressing bytes, 2 null bytes
+#define LOR_HEARTBEAT_SIZE           5
+#define LOR_ENHANCED_CMD             0x8E
+
+/////////////////////////////////////////////////////////////////////////////
+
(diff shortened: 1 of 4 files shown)


davidcorrigan714:
Oh, not sure I got the module building in the full build - I was invoking the module mk file directly and wasn't sure if that includes it in a clean build.

dkulp:
Very cool.  Thanks!

ryanspiffy:
@davidcorrigan714 I am trying to use LOR Enhanced this year for a few remaining LOR-based props that I have.  Last year I had them running on the LOR output from FPP, but they were a little slow to update, so I'm super excited to test out the 500K/1M speed option.  I couldn't find the additional documentation on the LOR Enhanced setup; did it make it into the guide?  If not, can you explain what the FPP Start Address on each Unit is in relation to the Start Channel set for the LOR Enhanced row?

darylc:
@ryanspiffy You may want to consider a pixielink device, that lets the Pixies run at even higher speeds - 10M from memory and as a bonus lets FPP use e1.31/ethernet instead of USB dongles.

ryanspiffy:
@davidcorrigan714 I'll have to look into that Pixielink, but most likely I'll be replacing all these with e1.31 stuff next year. My props are using the older Cosmic Color Bulbs that were two 50 bulb strands per controller.  I actually have managed to get it working with the new LOR Enhanced protocol that you added, so thanks for that!  I did some testing on one of the CCBs and am getting much faster refresh rates now.  Just a lot of trial and error to figure out the right combination of settings to address all 4 CCBs I have at the same time.  They are all run in series from the USB dongle and were running on LOR regular network from FPP, so just trying to get them working the same on LOR Enhanced.

davidcorrigan714:
The Pixielink looks pretty nifty and really not expensive for what it is. I never tried it with the CCB stuff though I'd imagine the first port was just the ID set on the controller then port 2 was the next one? Don't think I have an FPP on hand that's actually flashed with that code at the moment so hopefully it made enough sense with the start addresses once you figured it out.

With the Pixielink stuff supporting faster data rates I was really curious what the range on them would be, and if it's the same enhanced protocol or something different. It'd be trivial to bump up the baud rates in the code if you wanted to try to go even faster. Since it's the same physical chips & wiring it's not like the protocol can really much to send more data without compromising something, like bit depth on the colors. The Enhanced protocol is simple enough as it is and I personally am not a fan of trying to compress things - keeping it simple and hard real time at the sequence refresh rate has always been my preference.

ryanspiffy:
Well I was making good progress with it; but I can't seem to get downstream CCB units 2, 3 and 4 to work.  Doing more trial and error.  So far, I've noticed that setting the LOR Unit ID doesn't seem to have any effect on what unit I'm addressing.  I'll do some more testing tomorrow.

davidcorrigan714:
Looks like there might be options to have each port take a unit ID, or do 2 unit IDs per board. For having 2 unit IDs per board I would've thought that something like this would work:

image

For 1 ID per board it'd probably just be 4 Unit Ids, then 100 pixels each.

ryanspiffy:
So that was my issue; it had to do with how the CCBs were configured.  While you can set them up for a Single ID each; doing so causes the second string to get "off key" for the RGB channels.  The reason is the macro channels; channels 1-150 are RGB, and 151-160 are macro effects (I don't use these but they still map to channels).  This means that the second string in Single ID mode starts on channel 161.  Since the LOR Start is using a Pixel ID rather than channel number, you can't get the second string to line up correctly so that RGB falls onto the correct channels.  You could start the second string on LOR Start Pix 54/55 but that would start them on channel 159 or 162 when it needed to start on 160.  Might be nice to have the option to set the LOR Start Channel instead for future compatibility.

The workaround/fix for me was to remap all the CCBs to Dual Unit ID mode (1/2, 3/4, 5/6, 7/8), and then just use the first 50 pixels (150 channels) of each new Unit ID.  Everything is working correctly now, and much faster than it was using the LOR option last year.  Thanks again for adding this!

PR #947, opened 2021-01-05
start looping effects from mqtt

thekev:
Topic `{baseTopic}/set/effect/start` will play an effect once. This PR adds `effect/start/loop` to implement looping effects via MQTT.

### DIFF
--- a/src/mqtt.cpp
+++ b/src/mqtt.cpp
@@ -530,6 +530,8 @@ void MosquittoClient::MessageCallback(void *obj, const struct mosquitto_message 
 
 		} else if (topic == "effect/start") {
 			StartEffect(payload.c_str(), 0);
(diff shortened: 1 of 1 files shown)


cpinkham:
The /set/effect topic path is a legacy deprecated topic.  FPP Commands may be triggered via MQTT now and that is the preferred method to start effects now.

The topic is "{basetopic}/set/command/Effect Start" and the payload is the JSON formatted data for the arguments for the FPP Command.  You can look up the required arguments by looking through the FPP Commands at http://YourFPPSystemIP/api/commands

The Effect Start command takes 4 args:
effect = the effect name
startChannel = integer start channel override
loop = boolean true/false
bg = boolean true/false (run as a background effect)

We may remove the deprecated /set/effect support at some point, but it is not currently on the ToDo list.

PR #966, opened 2021-02-16
fix docker-compose.yml

jnovack:
* ensured that `npm install` is run after directory is mounted (this builds on first run, and 2-3 seconds to each subsequent run)
* rebuilds node-sass to avoid incorrect compilation on first-run (https://github.com/sass/node-sass/issues/1651) 
* exposed ports for BrowserSync
* runs `gulp watch-bs` for BrowerSync development

### DIFF
--- a/Docker/Dockerfile.node
+++ b/Docker/Dockerfile.node
@@ -1,5 +1,6 @@
 FROM node:latest
 
 WORKDIR /opt/fpp/www
-RUN npm install -g gulp ; \
-    npm install
+EXPOSE 3180 3181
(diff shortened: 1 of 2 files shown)


jnovack:
The first run may take up to 2 minutes depending on the processor you have.  Do you think that I should add `--verbose` to the `npm install` to let people know it's working?

Perhaps more logging between commands?

dkulp:
Considering how long it takes to build the main docker container, I'm not sure I'd worry too much.  :)

PR #968, opened 2021-02-16
Settings Grid 

jnovack:
To begin with Grid transition, only the Settings pages were updated.  I'm hoping with a successful review, I can move on to other pages to make consistent.

**Added**
* bootstrap `container`s and `row`s to the setting pages.

**Changed**
* `span` to `div` for settings descriptions so they respect `margin-top`
* halved padding for buttons globally.  buttons are ~20% smaller across interface.
* made ClockDate and ClockTime input widths consistent (that was bothering me)
* Current Time is now in the Time Config table to benefit from styling.

**Removed**
* `table`s
* colons after description

image

### DIFF
--- a/www/common.php
+++ b/www/common.php
@@ -221,7 +221,7 @@ function PrintSetting($setting, $callback = '', $options = Array(), $plugin = ''
         LoadPluginSettingInfos($plugin);
 
     if (!isset($settingInfos[$setting])) {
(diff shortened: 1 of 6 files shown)


adamcoulombe:
The approach seems right to me. I haven’t tested this in browser, but The only comment I have (and I think others will bring this up) is that the inputs are  a bit far from the field descriptions on the left which might make it difficult to follow from side to side , so maybe a smaller column or the left side?

PR #981, opened 2021-03-06
Prevent users bypassing basic auth on localhost when enabled by the owner

bendudz:
Added fine grained auth after a feature was discovered when users are utilising a wormhole (ie https://www.dataplicity.com). This branch will allow local fpp requests to the internal api to continue to work but also prevents anyone on the local system bypassing basic auth if it is enabled by the system owner.

### DIFF
--- a/www/api/.htaccess
+++ b/www/api/.htaccess
@@ -1,3 +1,26 @@
+SetEnvIf LOCAL_PROTECT 1 PROTECTED
+
+AuthUserFile /home/fpp/media/config/.htpasswd
+AuthType Basic
+AuthName "Falcon Player"
+
+<RequireAny>
+    <RequireAll>
+        # If parent is not protected - allow access
+        Require not env PROTECTED
+        Require all granted
+    </RequireAll>
+    <RequireAll>
+        # If parent is protected & request comes from local - allow access
+       Require env PROTECTED
+       Require local
+    </RequireAll>
+    <RequireAll>
+        # If user is authenticated - allow access
+        Require valid-user
+    </RequireAll>
+</RequireAny>
+
 RewriteEngine on
 
 RewriteBase /api/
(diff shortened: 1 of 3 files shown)


www/api/.htaccess
```diff
@@ -1,3 +1,6 @@
+Require local
+Require valid-user
+
```
dkulp, line 3: This completely breaks normal usage of FPP where a password is not configured.

bendudz: Ah I see. Let me rethink this then. 

Would you accept a PR if I switch this feature on using the SetupHtaccess function?

dkulp: It cannot modify any file in /opt/fpp as it would get wiped out on an upgrade. 

Also note that there are a bunch of calls to fppjson and fppxml from the parts that have to work via localhost.  For example, the OLED based menus on the various controllers currently make calls via fppjson and fppxml.  Until those are changed over to API calls, I don't think we can restrict localhost.

bendudz: Hi Dan, I've pushed a new revision of the changes.

I've modified the .htaccess in www/api to add auth depending on the system configuration with no password / password & external / internal access. 

common/settings.php has been added to. I've specifically called out allow any access to fppjson & fppxml as you said they will be deprecated as & when to move to /api. 

I have tested this in a docker environment & applied the changes on top of 4.6, burned a sd card & tried them on a Raspberry Pi & my tests work as I would expect.

Hopefully this revision gets the green light. Thanks.

bendudz:
@dkulp did you get chance to review this change yet please?

ghormann:
I tested this on my local box.   It appears to work well, so I will merge it.   
1) When password is set, api/ and fppdjson.php and fppxml.php still work from localhost w/o a password
2) When password is set, other http pages are blocked 
3) When password is removed using the UI, the system works as expected. No password required to access anything.

PR #1010, opened 2021-05-31
Add "If Not Running" functionality to local Effect Sequences

darylc:


ghormann (approved):
Tested locally and works as expected.   Documentation is automatically updated on API Help Page.

PR #1012, opened 2021-06-01
Implement Remote playlist start command

darylc:


dkulp:
Can I ask why this is needed?   The normal "Start Playlist" command should already work for this.  If you turn on the multicast checkbox and then select the target remote, it should accomplish the same thing.

darylc:
@dkulp my use case for the changes I've made in the past few days is to have one standalone fpp cause things to happen on another standalone fpp.  Think more along the lines of having a remote start a playlist on a master. (All the commands are called Start xyz on Remote, but they work for any FPP IP address you set them to)

dkulp:
Right, but every single FPP Command can already act that way.   We don't need to create new commands for that as they all already do that.   We have a few "Remote Start" type commands, but those are mostly legacy from before commands were able to be multicast out and likely should be removed, not more added.

darylc:
Multicasting does not work to different subnets.

darylc:
@dkulp You know neither the big buttons config page nor the API help reference the multicast functionality, but I see what you mean now - you can enter host IP addresses when selecting multicast when setting it up under command presets.  I thought you meant IP multicast not a custom FPP multicast in your last comment.

darylc:
@dkulp After testing and looking into this further, I'd request that the "Remote start" type commands are not removed without further improvements to the Multicast functionality.  It does not work across different IP subnets like the current "Remote start" type commands do.  Thus this pull request is relevant until such time as that restriction is removed.

ghormann (approved):
Works well in my testing.

PR #1016, opened 2021-06-05
Add descriptions to GPIO input ports that can be entered by users

darylc:


ghormann:
Tested and looks good. 

This page needs overhauled to replace tables and be a bit more dynamic.

PR #1025, opened 2021-06-29
Feature - Allow layering of fseq effects by skipping 0 values.

futileventure:
I'm looking for some feedback on this. I could see this being an effect option alongside background and loop to change this, however, I'm not a big fan of passing conditional flags into a methods arguments. Beyond that, I could see different colors being used as the "alpha mask" (more or less). Thoughts?

### Example : 

explanation

### DIFF
--- a/src/fseq/FSEQFile.cpp
+++ b/src/fseq/FSEQFile.cpp
@@ -629,6 +629,7 @@ V1FSEQFile::V1FSEQFile(const std::string &fn, FILE *file, const std::vector<uint
 V1FSEQFile::~V1FSEQFile() {
 }
 
+
 class UncompressedFrameData : public FSEQFile::FrameData {
 public:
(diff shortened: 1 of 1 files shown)


cpinkham:
At a minimum I think this would need to copy based on RGB triplets and that could get complicated real fast.  The Pixel Overlay code can act in a transparent mode such as you have in your patch where it checks each individual channel but it can also act in what we called Transparent RGB where it looks at RGB triplets and sees if any are non-zero.  You probably don't want the a pixel showing the blue channel from one sequence and the red and green channels from a different sequence because the colors would be all out of wack.

Thoughts?

cpinkham:
Closing for now since no reply.  If you would like to discuss further, bring it up on the FalconChristmas.com forum and we'll see if we can come up with a workable solution.

PR #1026, opened 2021-06-30
Ping IP XSS fix

x3rz:
Sanitized the ip parameter of ping.php which was responsible for reflected XSS which is [reported](https://www.huntr.dev/bounties/1625036661877-FalconChristmas/fpp/) on huntr.dev


ghormann:
Thank you for the PR

PR #1037, opened 2021-07-17
Falcon v4

darylc:
Add Falcon v4 information to multisync page


www/multisync.php
```diff
@@ -844,6 +855,58 @@ function getFalconControllerStatus(ip) {
     });
 }
 
+function getFalconV4ControllerStatus(ip) {
+$.ajax({
+        url: '/proxy/' + ip + '/api',
+            method: 'POST',
```
ghormann, line 861: This is a post call? Interesting.....  I don't have one to test with, so I assume it is.   Just seemed odd.

ghormann (approved):

www/multisync.php
darylc: Yes definitely a post, the API requires you to tell it that you want to receive the status v's asking the API to do something for you.

PR #1043, opened 2021-07-21
Move RGB order message to the RGB Order section.

Poporacer:
The Informational Banner about the RGB order wasn't in the section that it applied to and wasn't prominent. Moving it to this location might be a little more obvious?


ghormann (approved):
I agree, this is better.

PR #1045, opened 2021-07-23
Detect if a proxy is needed to access the Falcon v4's

darylc:


ghormann (approved):

patdelaney:
@darylc Can you describe what this change is for and if it has any implications for FPP Documentation?

darylc:
No FPP documentation changes.  This is just a change to make the v4 controllers work properly on the multisync page when your web browser does not have direct ip connectivity to the v4 controller.

PR #1047, opened 2021-07-25
Add a privacy settings tab

darylc:
Moved the fppstats to this tab
Add settings to turn off phone home to cape vendors

### DIFF
--- a/www/settings.json
+++ b/www/settings.json
@@ -95,6 +95,8 @@
                 "HostName",
                 "HostDescription",
                 "statsPublish",
+                "FetchVendorLogos",
+                "SendVendorSerial",
                 "Locale",
                 "TimeZone",
                 "Latitude",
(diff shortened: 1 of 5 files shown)


www/settings.json
```diff
@@ -627,6 +636,17 @@
             "level": 1,
             "type": "checkbox"
         },
+	"FetchVendorLogos": {
+            "name": "FetchVendorLogos",
+            "description": "Fetch cape logos from vendors",
+            "gatherStats" : true,
+            "tip": "Some capes fetch the vendor logo from their website when using the FPP gui.  This discloses your IP address to the vendor.",
```
ghormann, line 643:
```suggestion
            "tip": "Some capes fetch the vendor logo from their website when using the FPP UI.  This discloses your IP address to the vendor.",
```
Nit pick change.

ghormann (commented):
@darylc  I think the same setting for cape vendors should also affect the collection of cape UUIDs for stats.  

My proposal that this new setting (SendVendorSerial) control if  'serialNumber' is gathered in this section is well.  Can you make this addition?

https://github.com/FalconChristmas/fpp/blob/master/www/api/controllers/stats.php#L387-L393

Issue #1081, opened 2021-08-31
Add Panel Size Scan Rates for P4 ColorLight card Panels


bendudz:
**Is your feature request related to a problem? Please describe.**
Greg Macaree of https://panelsrus.co.uk has requested support for P4 panels for a particular size & scan rate.

**Describe the solution you'd like**
An option of 80x40x10 & 80x40x20 single panel sizes when a colorlight card is selected on the LED Panels tab.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

darylc:
Greg also said he would send a pull request for this, so I am a little confused as to why you've logged this as an issue?

bendudz:
> Greg also said he would send a pull request for this, so I am a little confused as to why you've logged this as an issue?

Sorry - I'd spoken to Greg about this & he was happy for me to add the functionality. I've raised a PR https://github.com/FalconChristmas/fpp/pull/1082 to add this.

darylc:
ok lets close this issue now that the PR is accepted

PR #1082, opened 2021-08-31
Added extra panel sizes to support P4 on ColorLight cards

bendudz:
Changed how panel sizes are loaded to allow specific sizes & scan rates based on the connection field.

https://github.com/FalconChristmas/fpp/issues/1081

### DIFF
--- a/www/co-ledPanels.php
+++ b/www/co-ledPanels.php
@@ -452,6 +452,7 @@ if ($settings['Platform'] == "Raspberry Pi" || $settings['Platform'] == "BeagleB
 }
 ?>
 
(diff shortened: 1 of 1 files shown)


darylc (commented):
I would have thought support for these panel sizes should also be added for LINSN not just colorlight

bendudz:
> I would have thought support for these panel sizes should also be added for LINSN not just colorlight

Added Linsn too

dkulp:
Would you guys happen to have a Octoscroller or PocketScroller available?  I'm 90% certain that the Beagles could support these as well, but don't want to just add them to the list without testing.

PR #1164, opened 2021-11-23
PHP PSR-12 Extended Coding Style

mattwells:
I have gone through the PHP files an updated them as best I can to conform to the [PHP PSR-12 Extended Coding Style ](https://www.php-fig.org/psr/psr-12/) as best that I can.

There should be no changes to the functionality of the code as they are visual code changes only. My intention is to begin refactoring the code to make it easier to convert to a framework such as Laravel or Symfony in the future, as time permits.

Using [PHP CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) Code Beautifier most of it was done automatically with the final report looking like this:

```
$ phpcs --standard=PSR12 -ln .

FILE: C:\Users\Matt\source\repos\fpp\www\cronjobs.php
-----------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
-----------------------------------------------------------------------------------------------
 3 | ERROR | Each class must be in a namespace of at least one level (a top-level vendor name)
-----------------------------------------------------------------------------------------------


FILE: C:\Users\Matt\source\repos\fpp\www\pixelnetdmxentry.php
-----------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
-----------------------------------------------------------------------------------------------
 3 | ERROR | Each class must be in a namespace of at least one level (a top-level vendor name)
-----------------------------------------------------------------------------------------------


FILE: C:\Users\Matt\source\repos\fpp\www\scheduleentry.php
-----------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
-----------------------------------------------------------------------------------------------
 3 | ERROR | Each class must be in a namespace of at least one level (a top-level vendor name)
-----------------------------------------------------------------------------------------------


FILE: C:\Users\Matt\source\repos\fpp\www\universeentry.php
-----------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
-----------------------------------------------------------------------------------------------
 3 | ERROR | Each class must be in a namespace of at least one level (a top-level vendor name)
-----------------------------------------------------------------------------------------------

Time: 757ms; Memory: 20MB
```

I have added the `n` flag which ignores warnings, the `l` flag stops it going through directories recusively as there is a `lib` directory which I'm unsure if it comes from a third party or not.

The final errors are to do with classes and until some form of autoloading is setup I feel it is best to leave those classes as they are for now.


mattwells:
Is there an interest for this? I understand if there is not but I am be hesitent to do more without feedback.

dkulp:
There are a few parts to this question...

For the most part, cleaning up the PHP code is definitely worth doing.   Some of that code is very old and is using tabs and such which does cause some headaches while editing.   Simple cleanup to make the code more consistent would be welcome, but it would need to be in small parts.  Reviewing 70+ files at once is not something I really have any interest in doing.

Second, whatever cleanup is done would need to match the defaults for what happens when editing the code via vscode and the PHP configurations and such that it uses.   There is no sense cleaning up code into one format that will be completely changed again the first time a developer touches the file.   At this point, vscode setup as a remote (see the FPP-with-vs-code.md file in the docs subdirectory) is likely the most used way of editing the code so that needs to be the primary configuration that has to "just work".

Finally, you mention converting to a framework like Laravel or Symfony.  That is most likely a non-starter.   The developers talked in depth about that last year and pretty much decided that that is undesirable at this point.   In the last year, we pretty much re-wrote much of the UI.  We are definitely not going to do that kind of undertaking yet again any time soon.   If the primary goal of your efforts are to move to a different framework, then it's likely a waste of your time to continue.

mattwells:
Yes I understand and agree that that many files is a lot, I will start seperating it into smaller chunks so they can be reviewed in smaller chunks.

As for VS code that is what I was using as the editor, I didn't check what the PHP formatter would do by default though but I will check. I would assume that Microsoft (or the addon developer) would have followed a standard of some type so I would assume it would be reasonable to go into the file and use the addon formatter rather than the tool I was using.

As for moving to a framework, I couldn't imagine that moving to a framework would be a cause for concern. This would be something down the road though after the codebase has been modernized. 

There is a book called [Modernizing Legacy Applications In PHP by Paul M. Jones](https://leanpub.com/mlaphp) (a well know developer in the Laravel community) which I would be loosely following to update the code base but in general this is the kind of app structure this book is designed for.

PR #1202, opened 2021-12-12
Add new feature to Extend Schedule API. 

darylc:
You can now specify an optional 2nd parameter Limit - a number of seconds that will cause FPP to extend start/end times for all future schedules that fall within the next Limit seconds.

api/command/Extend Schedule/SECONDS[/LIMIT]

eg api/command/Extend Schedule/600/3600 will extend all start/end times due to occur within the next 1 hour by 10 minutes

Thanks to Captain Murdoch for his assistance with this feature.


darylc:
replaced by https://github.com/FalconChristmas/fpp/pull/1203

PR #1203, opened 2021-12-12
Add new feature to Extend Schedule API. 

darylc:
You can now specify an optional 2nd parameter Limit - a number of seconds that will cause FPP to extend start/end times for all future schedules that fall within the next Limit seconds.

api/command/Extend Schedule/SECONDS[/LIMIT]

eg api/command/Extend Schedule/600/3600 will extend all start/end times due to occur within the next 1 hour by 10 minutes

Thanks to Captain Murdoch for his assistance with this feature.

### DIFF
--- a/src/ScheduleEntry.cpp
+++ b/src/ScheduleEntry.cpp
@@ -296,7 +296,7 @@ void ScheduleEntry::pushStartEndTimes(int dow, int &delta, int deltaThreshold) {
     if (delta != 0) {
         // Only adjust if the threshold time is in the future
 	if (deltaThreshold > time(NULL)) {
-            if (deltaThreshold > startTime) {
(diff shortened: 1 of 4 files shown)


cpinkham:
I think this still needs one more change.  The "delta = 0" needs to be in the else portion of the "if (deltaThreshold > time(NULL)) {" statement, otherwise delta will get reset to 0 the first time pushStartEndTimes() is called with an end time that is less than deltaThreshold.  We want to keep delta set until after the deltaThreshold end time has passed, otherwise if the scheduler runs twice before the deltaThreshold time, the second time it runs delta would have been reset to 0.

darylc:
replaced by https://github.com/FalconChristmas/fpp/pull/1235

PR #1208, opened 2021-12-16
Robots file to prevent FPP being listed by Google

bendudz:
It won't prevent mallicious bots / actors from accessing the FPP instance, but should stop Google & other 'friendly' bots from scraping it.


ghormann:
Not sure if it will really do any good, but  can't hurt. :-)

PR #1234, opened 2022-01-20
sort the list of Other output types (to be more user-friendly), add troubleshooting "back to top"

djulien:
- might want to apply sorting to other lists as well


cpinkham:
I think we said we could close this now since these were covered by other commits and #1239.

PR #1244, opened 2022-02-10
Show FPS and slowest port for DPI Channel Outputs

djulien:
Since there are so many lines displayed on the DPI 24 page, it's helpful to know which port has the slowest refresh rate.
Also, added example config line to show how to force DPI to fb1, and cross referenced GPIO pin#s within .cpp file.

### DIFF
--- a/www/co-pixelStrings.php
+++ b/www/co-pixelStrings.php
@@ -635,6 +635,9 @@ function getPixelStringOutputJSON()
 
 		output.outputs = outputs;
 		postData.channelOutputs.push(output);
+//                if (IsPixelStringDriverType(output.type)) {
+                const is_dpi = ~output.type.indexOf("DPI");
+                show_slowest(output, is_dpi); //show extra DPI info (fps for slowest port)
     });
 
     return postData;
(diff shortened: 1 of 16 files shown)


www/co-pixelStrings.php
```diff
@@ -1362,6 +1367,61 @@ function ValidateBBBStrings(data) {
...
+
+    16: "GPIO20 (P1-38)",
+    17: "GPIO21 (P1-40)",
+    18: "GPIO22 (P1-15)",
+    19: "GPIO23 (P1-16)",
+    20: "GPIO24 (P1-18)",
+    21: "GPIO25 (P1-22)",
+    22: "GPIO26 (P1-37)",
+    23: "GPIO27 (P1-13)",
+};
```
dkulp, line 1398: This information needs to be pulled from the cape JSON.   Different capes/hats can use different pins for each port and thus this map would not be correct.

djulien: The "GPIO" piece is not in the capes.  Is it okay if I add that into the pin names in the cape .sjon file?

dkulp: No, just the pin names in there....   For now, I would change this to a map like:

```
const PIN2GPIO = {
  "P1-7": "GPIO4",
  "P1-29: "GPIO5",
..
}
```
and use that to create the needed string.   Ideally, even this would be pulled via the GPIO rest API: 

http://fpp/api/gpio

cpinkham:
I am looking into the ptr overrun issue, I can reproduce it with Don's co-pixelStrings.json.  I have the hdmi force setting and GPIO pin comments updated as part of a larger patch I am mostly done with that adds support for smart receivers and latches.  I need to look at the FPS code a bit more.  Don and I chatted briefly about that as well but I haven't had a chance to look it over yet.

www/co-pixelStrings.php
djulien: Okay, I'll change it to use /api/gpio.  I didn't see that one listed in REST api help, but it gives exactly the info I wanted.  thanks!

djulien:
closing this PR and creating a new one to avoid extraneous commited files

PR #1264, opened 2022-03-30
disable table header non-scroll and hide Import button

djulien:


djulien:
latest change is to only show Import button if xlights_rgbeffects.xml file is in uploads folder.

PR #1295, opened 2022-06-22
Remove + from offset since you can have -30

cybercop23:
If you want to start x min before SunSet/SunRise etc... you can just enter - offset minutes.  Remove the + since that's ambiguous.


cybercop23:
@cpinkham I was going to put in a request to allow for negative offset, but then I tested it and it worked. So best to remove the + and document that - offset is allowed. 
FYI @Poporacer

cybercop23:
Or if desired make it +/-

PR #1356, opened 2022-09-07
enh hostname validation

cybercop23:
Enhance hostname validation so foo.local and foo.bar.local are valid


Poporacer:
I don't think foo.local is a valid HostName, it is a domain name.

PR #1364, opened 2022-09-21
Fix free OS check issue

cybercop23:
fix in 6.1 branch so users can upgrade before next master realease

### DIFF
--- a/www/about.php
+++ b/www/about.php
@@ -257,7 +257,7 @@ function OSSelectChanged() {
     var os = $('#OSSelect').val();
 <?
 // we want at least a 200MB in order to be able to apply the fppos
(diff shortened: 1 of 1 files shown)


Poporacer:
This was already implemented in the Master Branch

cybercop23:
Yes, but this is needed so users that are on 6.1 can upgrade OS from the About menu. Right now they are stuck and won't be able to ever use that. They'l have to manually download and install via the Content/File Manager.

Poporacer:
The next release, 6.1.1 will be a minor release which will not require an OS upgrade, just a normal FPP Upgrade and will fix the OS Upgrade button issue.

Poporacer:
Implemented in version 6.1.1

PR #1463, opened 2023-01-21
Automatic JSON Settings Backup & Store In Alternate Location (optional)

### LINKED ISSUE #1148: Incorporate automatic backups
My SD card was fried today and I was saved because I had a current backup.  But others may not have current backups or even know that backups are possible.  I suggest making an automatic backup after changes are made.  The backups are essentially tiny and would be very helpful during crunch time when they are most needed. 

An added feature would be to delete older backups, say 90 days, to conserve space if needed.

jaredb7:
The current implementation generates a backup after the setting change is made for specific API endpoints or actions (listed below).
The required API endpoints have been modified to make a function call to generate a backup after those respective settings are applied/committed.

In addition, Users can manage the JSON backups residing on their device, remove backups if they don't need them, download backups for safe keeping. If deleted backup exists on SD and USB it is deleted from both locations.
Restoration of any backup on the device can also be carried out through a new section under the restore section

A option to copy these setting backups to USB has also been implemented, The user will need to specify which USB storage device is to be used via a setting on the FPP Backup page. After which, all existing setting backups on the device will be cloned to the chosen device and all future backups will be copied there also using the copy_settings_to_storage script.

Areas Covered by in the automatic settings backup

> **Settings**
> - Any Setting Changes
> - Schedule Updated
> 
> **Playlist Changes**
> - Playlist Created
> - Playlist Updated
> - Playlist Deleted
> - Playlist Content Modified
> 
> **Input / Output Setup**
> - Output Configuration Change (Channel Input / Output)
> - Output Processors change
> - Pixelnet DMX Configuration Change
> - Config File Uploaded (Proxies, GPIO, Pixel Overlay changes)
> 
> **System Changes**
> - Email Configuration Changed
> - DNS Configuration changed
> - Script Installed
> - Script Modified

Closes #1148 

image
image
Prompt to copy to USB when the storage device is set
image

### DIFF
--- a/scripts/copy_settings_to_storage.sh
+++ b/scripts/copy_settings_to_storage.sh
@@ -118,6 +118,15 @@ for action in $@; do
     "Backups")
         rsync $EXTRA_ARGS $SOURCE/backups $DEST  2>&1 | IgnoreWarnings
         ;;
+    "JsonBackups")
+        if [ ! -d "$DEST/config/backups" ]
+        then
+            mkdir -p $DEST/config/backups 2> /dev/null
+            chown fpp.fpp $DEST/config/backups 2> /dev/null
+            chmod 775 $DEST/config/backups 2> /dev/null
+        fi
+        rsync $EXTRA_ARGS $SOURCE/config/backups $DEST/config  2>&1 | IgnoreWarnings
+        ;;
     "Configuration")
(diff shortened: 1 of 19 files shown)


jaredb7:
Leaving this PR to sit in case something comes up. Tested everything + reflash & restore backup from USB

PR #1475, opened 2023-02-05
updated checkTimeAgainstUSNO to match /opt/fpp/scripts/functions

OnlineDynamic:
Updated copy of function in FPP_install.sh to match /opt/fpp/scripts/functions

### DIFF
--- a/SD/FPP_Install.sh
+++ b/SD/FPP_Install.sh
@@ -97,8 +97,13 @@ checkTimeAgainstUSNO () {
 
 		# allow clocks to differ by 24 hours to handle time zone differences
 		THRESHOLD=86400
-		USNODATE=$(curl -v --silent https://nist.time.gov/ 2>&1 \ | grep '< Date' | sed -e 's/< Date: //')
+        USNODATE=$(curl -I --connect-timeout 10 --max-time 10 --retry 2 --retry-max-time 10 -f -v --silent https://nist.time.gov/ 2>&1 \ | grep '< Date' | sed -e 's/< Date: //')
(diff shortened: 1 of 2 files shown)


SD/FPP_Install.sh
```diff
@@ -97,7 +97,8 @@ checkTimeAgainstUSNO () {
 
 		# allow clocks to differ by 24 hours to handle time zone differences
 		THRESHOLD=86400
-		USNOSECS=$(wget --timeout=10 --tries=1 -q -O - http://www.usno.navy.mil/cgi-bin/time.pl | sed -e "s/.*\">//" -e "s/<\/t.*//" -e "s/...$//")
+		USNODATE=$(curl -v --silent https://nist.time.gov/ 2>&1 \ | grep '< Date' | sed -e 's/< Date: //')
+        USNOSECS=$(date -d "${USNODATE}" +%s)
```
ghormann, line 101: While fixing this, wouldn't it be better to add some error handling?     It is possible for line 100 to return a 404 and it still then attempts to call `date -d` .  This happens in the event of a DNS failure.

Also, the lack of  a timeout  seems wrong.      I thought by default curl had a reasonable timeout but I just did some testing and that wasn't the case....    :-(

ghormann (changes requested):

SD/FPP_Install.sh
OnlineDynamic: Good point and easy fix will change the pull request when back at a pc

ghormann (approved):

PR #1477, opened 2023-02-06
Added gpio info to troubleshooting commands

OnlineDynamic:
Addition of useful info when wanting to check status of internal and external gpio lines and devices


dkulp:
Can you make sure you rebase against master before submitting a pull request so only your changes are shown?   In general, a "git pull -r origin/master" will do it.

I believe these checks should be excluded on macOS as there isn't gpiodetect and gpioinfo on that platform.

OnlineDynamic:
will close and reopen with a cleanedup version which excluded command on macOS

PR #1478, opened 2023-02-06
Addition of GPIO detection to troubleshoot commands

OnlineDynamic:
Addition of gpiodetect and gpioinfo commands on non macOS based platforms


OnlineDynamic:
@darylc added extra code to check for presence of the commands

darylc (approved):

PR #1479, opened 2023-02-07
File Copy Backup To Remote Host's USB Device

### LINKED ISSUE #778: Enhancement Request - Select destination for remote backup
I would like to have the ability to select the destination for remote backups.
Example:
I have an external USB 512gb SSD on the master FPP. 
I would like to have a selection of what drive to save backups to on the master FPP.
This away I could steer to the SSD (or other USB drive) when using "Backup to remote" on a remote FPP instance instead of the SD card which is rapidly getting very full.

jaredb7:
This was a tricky one and only really made possible by the [nsenter ](https://man7.org/linux/man-pages/man1/nsenter.1.html) command which allows us to run commands in a different process's namespace.
This enables us to mount a device (under the root process) and make it available to the entire system, since apache sandbox's or more specifically mounts created in the apache process are private to apache.

I believe this command is available by default in most linux installation so shouldn't pose any issues.
Works on Raspbian, just needs to be tested on BBB but I gather it should work  there also

The rSync Config is also updated (via upgrade script 76) with a new module [remote_filecopy] which points to a predetermined and expected mount point at /mnt/remote_filecopy (specified USB device is mounted here)

**Backups API**
Added 2 endpoints which could be called to mount a specific device to a specified directory, useful only if a service on the remote device is expecting that mount location.
And Unmount that same location
Updated DriveMountHelper with ability to specify mount location,
-whether mount location is unmounted once done,
-whether to perform the mount in the root mount namespace (uses nsenter)
-or to return result codes etc for all the commands run (useful for debugging)

**Backup to Remote Host**
The File Copy Backup page To Remote and From remote options receive a extra dropdown box which is the available USB devices on that host,
image

**Restore from Remote** Host, leverage existing code to get list folders on the specified/selected device
image

**Copy Process**
The copystore.php page gets a slight change to allow passing of the &remoteStorage query string (if omitted, we substitute in 'none' to the script.
image

**Copy Settings Shell Script** - copy_settings_to_storage.sh
The shell script received changes to the params passed to and the TOREMOTE and FROMREMOTE sections to call a API to mount the specified device on the remote host IF a device is specified, Rsync continues as normal as the new module is configured pointing to the expected mount point.
Device is unmounted at the end

Closes #778

### DIFF
--- a/etc/rsyncd.conf
+++ b/etc/rsyncd.conf
@@ -10,3 +10,10 @@ uid = FPPUSER
 gid = FPPUSER
 hosts allow = 127.0.0.0/8 172.16.0.0/12 192.168.0.0/16 10.0.0.0/8
 
+[remote_filecopy]
+path = /mnt/remote_filecopy
+read only = no
+list = no
+uid = FPPUSER
+gid = FPPUSER
+hosts allow = 127.0.0.0/8 172.16.0.0/12 192.168.0.0/16 10.0.0.0/8
--- a/scripts/copy_settings_to_storage.sh
+++ b/scripts/copy_settings_to_storage.sh
@@ -1,9 +1,10 @@
 #!/bin/bash
 DEVICE=$1
-DPATH=$2
+DPATH=$2 # Folder For backups
 DIRECTION=$3
-DELETE=$4
-shift 4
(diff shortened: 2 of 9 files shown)


jaredb7:
Resolved merge conflict for upgrade 75, moved required changes into upgrade 76

PR #1502, opened 2023-04-15
Fix Website URL Markdown Formatting in README

LamboCreeper:
There is currently a space within the markdown formatting for the URL which points to the Falcon Player website in the README, therefore, causing it not to render correctly. This fixes it to align with the other links in the README.


ghormann (approved):

ghormann:
Thanks @LamboCreeper

PR #1543, opened 2023-07-21
Remove bold html formatting in rsync tip as it is not rendered

northrup:
The tips do not render html elements, as such this element displays incorrectly in the tip pop-up.

### DIFF
--- a/www/settings.json
+++ b/www/settings.json
@@ -1907,7 +1907,7 @@
             "name": "Service_rsync",
             "description": "Enable rsync",
             "gatherStats": true,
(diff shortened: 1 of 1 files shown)


dkulp:
This is likely irrelevant now.  I enabled the html formatting in the tool tips.  Can you test/check?

northrup:
🤔 I’m at the latest commit on `master`, sha 954a115, and I still see it rendering the same.

image

dkulp:
I suppose it helps if I push from the RIGHT beagle...  :(

northrup:
I’ve certainly done that myself, only generally discovering it after a moderate CI run thinking “why the heck aren’t my changes there?!”

This did indeed fix the html rendering 🎉

PR #1575, opened 2023-09-11
Hostname change on initial setup

cybercop23:
New SOP is to use hostnames in setting up ctrls/players and proxies, so this helps to keep them unique if they have more than one and one less change when we help them configure it (the right way).

### DIFF
--- a/www/initialSetup.php
+++ b/www/initialSetup.php
@@ -23,6 +23,11 @@ function finishSetup() {
             return;
     }
 
+    if (settings['HostName'] == 'FPP') {
+        alert('Please change the default host name (FPP) to a unique name. ie FPPMain, MainPlayer, Player etc...');
+        return;
+    }
+
 <? if ($showOSSecurity) { ?>
     if ($('#osPasswordEnable').val() == '') {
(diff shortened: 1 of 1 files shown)


darylc:
I am not in favor of this, users will forget what they set it to and fpp.local won't work.

cybercop23:
Well, the Host name and Desc is on the Inital Setup screen so (I would think) the intent is for user to change and customize it.
More and more users are running more than one FPP instance, controlers and/or projectos and they can't all be named FPP.
If they are that clumsy, there's always xLights -> FPP Connect to discover their newly named devices.
But again, can't have all of their FPP instances named FPP. What will that do?

darylc:
I've discussed this with you in zoom before and my stance hasn't changed.  This change is detrimental to novice FPP users who have a single FPP.  If users have more than 1 FPP they are free to change the hostname in the initial setup, but forcing it on others is a bad idea because it will prevent fpp.local from working and they will forget what they named it.

dkulp:
I'm also against this.   Many people that have a single FPP instance just keep the FPP name for simplicity.

PR #1577, opened 2023-09-12
Create PCF8574 support

alexbgreat:
This adds support for the PCF8574 i2c IO expander chip for output. Used in some relay boards and also an Adafruit GPIO expander. Supports simple (!= 0), threshold, and hysteresis output, as well as pin value and pin ordering inversion. I have tested this code on the relay board below.

Example relay board: https://www.amazon.com/Channel-Electromagnetic-Arduino-Raspberry-Indicators/dp/B083LFPCK8
GPIO Expander: https://learn.adafruit.com/adafruit-pcf8574/overview

### DIFF
--- a/src/channeloutput/PCF8574.cpp
+++ b/src/channeloutput/PCF8574.cpp
@@ -1,0 +1,242 @@
+/*
+ * This file is part of the Falcon Player (FPP) and is Copyright (C)
+ * 2013-2022 by the Falcon Player Developers.
+ *
+ * The Falcon Player (FPP) is free software, and is covered under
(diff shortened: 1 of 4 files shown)


dkulp:
Hmm... I'm not against this PR at all and I'll likely merge it once I get some time to look at it more.   That said, I kind of wonder if it would make some sense to somehow make it more "generic" by using the GPIO drivers built into the kernel and use the libgpiod API's.    That could then support a huge range of various GPIO devices including the 16bit chips instead of being hard coded to a single variant.    Something to think about....

Issue #1456, opened 2023-01-11
OLED Not working on RPI when GPIO input assigned to OLED Navigation Command


OnlineDynamic:
# Please answer the following:
**FPP Version and Hardware**: 
FPP 6.3 running on RPI Zero 2 W and RPI 4

**Describe the bug and Steps to reproduce**: 
I'm using the pca9554A gpio extender chip and an SSD1306 OLED.  Both work correctly togther on the I2C bus - extra GPIO's appear in the GPIO inputs fpp menu.  I can assign them to control volume and they are working correctly.  The OLED shows basic splash screen and ip address etc correctly.

However when I assign any of the GPIO's to 'OLED Navigation' command and restart RPI the OLED screen comes on with all its pixels lit up.

**Expected behavior**: 
OLED to work for FPP Menu navigation on RPI

**Additional context**: 
Thread on conversation here:
https://falconchristmas.com/forum/index.php?topic=15379.0.

**Additional Attachments**
Please include the log files. To create the file:
1. Open the FPP UI
2. Open the the File Manager (Content Setup->File Manager), 
3. Select the "Logs" tab
4. Click "zip" button.
5. Attach to this ticket.

[FPP_Logs_20230111.1420.zip](https://github.com/FalconChristmas/fpp/files/10393167/FPP_Logs_20230111.1420.zip)

darylc:
Seems unlikely that this will get looked at since none of us have the hardware.

OnlineDynamic:
Seem to have narrowed it down to line 196 in FPPOLEDUtils.cpp

`action->file = gpiodPin->line.event_get_fd();`

that is crashing the program with an error in debugger saying:

terminate called after throwing an instance of 'std::system_error'
  what():  unable to get the line event file descriptor: Operation not permitted

Any ideas before I invest lots of time researching this?

dkulp:
If I remember correctly (its been a long time since I looked), FPPOLED stuff doesn't support polling for GPIOs (just analog sensors) and using pca9554A for push buttons wouldn't work via interrupts unless FPPOLED is given the GPIO line that is used from the pca9554A for the interrupt.   The K16 and K32 both do that, but they don't configure the navigation via the gpio inputs page.    This will likely require a bit of work to support polling on the GPIO's, but it would also need to be done in a what that doesn't break the interrupt driven method.

darylc:
@OnlineDynamic Can this be closed?

OnlineDynamic:
[...]

PR #1590, opened 2023-10-04
Fix for #1456 plus some php code cleanup

OnlineDynamic:
#2 Commits in this pull request

1/ Please see #1456 for details - this introduces oled command into fppd which can talk to fppoled via shared memory to push actions across - minimum impact on fppoled so this is additional and backward compatible.

2/ clean up of functions not require in the API / Command help pages after they were seperated several months back (api one still had command related stuff and vice versa)


darylc:
I will defer to @dkulp for this PR, but I think @OnlineDynamic we should remove the debugging code left in gpio.cpp and cleanup the commented includes in FPPOLEDUtils.cpp?

PR #1598, opened 2023-10-09
Update rpi_ws281x for RPi 3B+ Rev 1.4 hardware.

clnorris:
I purchased both my rPI-28D+ and RPi 3B from Hansen Electronics. With that combo of boards, I saw this error message:

/opt/fpp/src/channeloutput/rpi_ws281x.cpp:208: ws2811_init() failed with error: -3

Digging deeper, the RPi's currently shipping from Hansen are RPi 3B+ Rev 1.4. rpi_ws281x needs a bump for a compatible out of box experience with rPI-28D+ and RPi purchased from Hansen.

Edit: I installed the v7.2 image and did a recompile with the change that I requested.  The Chase and Cycle RGB Test Patterns appear to be working.  Not sure if any other regression test is needed for rpi_ws281x.


clnorris:
@dkulp Thank you!

Issue #1609, opened 2023-10-16
Multisync Page Crash Report Link for Remote is not Absolute


MrPaulAR:
**FPP Version and Hardware**: 
Example: 7.2-2-ge35c0df7 on Pi4

**Describe the bug and Steps to reproduce**: 
I have a remote Pi that has 1 crash report.  The player correctly indicates that the crash report exists.  When I click on the crash report it goes back to the same device instead of the remote where the crash report actually exists.  This seems to be because the URL is reletave instead of absolute.

**Expected behavior**: 
Clicking on the link should take me to the IP of the remote.

**Additional context**: 
Here is a a screenshot showing the issue where the remote IP is 192.168.86.180 but the link to the crash report is going to 192.168.86.59 (self).  Not displayed on the screenshot is the location of the cursor which is directly above the crash report hyperlink.

![image](https://github.com/FalconChristmas/fpp/assets/3765696/ea36de4f-bff0-420c-8667-2ef5759cf2d6)

darylc:
Care to send a PR @MrPaulAR?

MrPaulAR:
Tried to figure out where the php was for this but couldn't find it.

If you want to spend time with me in zoom getting my IDE setup I'll be
happy to look at the php.

dartunder:
Couple hints.

Either fixed when the message is created, or in SystemGetStatus, or in multisync.php

fpp/www/api/controllers/system.php : function SystemGetStatus()
or 
fpp/www/mutlisync.php : around line 450 is where it's read then...
I think this is where it's set  $('#' + rowID + '_status').html(status);

darylc:
Hey Paul - you might find this video useful - https://www.youtube.com/watch?v=h2MhkFKUOow thats how I learned how to setup the IDE

MrPaulAR:
See comments in pull request.  In short, don't change the API output.  Issue is that the URL should resolve to proxied endpoints too.

PR #1610, opened 2023-10-17
Changed relative to absolute URL for crash reports

MrPaulAR:
Fixes https://github.com/FalconChristmas/fpp/issues/1609

### DIFF
--- a/www/api/controllers/system.php
+++ b/www/api/controllers/system.php
@@ -376,6 +376,7 @@ function finalizeStatusJson($obj)
         $num = count(glob($settings['mediaDirectory'] . "/crashes/*.zip"));
         if ($num > 0) {
             $plural = 's';
+            $ip = $_SERVER['HTTP_HOST'];
             $verb = 'are';
             if ($num == 1) {
                 $plural = '';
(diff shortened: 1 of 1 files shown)


dkulp:
If the FPP instance is behind a proxy FPP instance, the warning then displayed on the main status page would not work.   That's the main reason for using the relative URL.

MrPaulAR:
That makes sense on the API endpoint.

Going back to the issue for this PR, how can I address that?  I don't think
editing the multisync.php page will be aware of the proxy either (consider
I can run this on any FPP instance...not just the one doing the proxy).
Right now, the only time the crash url works in the multisync.php output is
if I'm on that device device.

dkulp:
My gut feeling is that it should be relative, but add something like class="localWarningLink" and then use some javascript or something on the multisync page to replace the URL's that have that class.

MrPaulAR:
JavaScript is beyond my current skillset that I'm comfortable showing to
others.  The PHP touched my old perl stuff...mostly python now.

I'm gonna close this pull request and maybe someone else will pick up the
issue.  It's cosmetic only so not a big deal if not.

PR #1679, opened 2023-11-29
Make links open in new tab

cybercop23:
Change proxy links and newly added hostname on multisync page open in new tabs


cpinkham:
I don't think it will cause any issues as-is, but can you add one more commit to wrap these target values in single quotes so it looks like    target='_blank' instead of   target=_blank  and the same for hostname?

PR #1681, opened 2023-11-30
Make hostname also open in new tab

cybercop23:


cpinkham:
Dan included this as part of bd9c4f9f538.  Closing this ticket.  Thanks for pointing it out.

PR #1682, opened 2023-11-30
Make Proxies Open in new tab

cybercop23:


cybercop23:
Hi @cpinkham. Made the changes suggested. Any other issues?

PR #1683, opened 2023-11-30
Added keyboard shortcuts to various pages

cybercop23:


cybercop23:
@dkulp @cpinkham is this something that you would consider merging? I got used to using they keys when I was creating this and would be really awesome to have.

dkulp:
I'm not sure "as is"...  Discussing it with other, I think if there was a modifier (CTRL-M or something) then it might be more acceptable.  Also, the leading / on the pages likely needs to be removed as that would keep things from working properly when behind an fpp proxy.

cybercop23:
@dkulp good point. Didn't think of proxies... I know folks have FPPs/Kulps behind a proxy. I can adjust that.
I would hate to add Ctrl.. although anything's possible... just love the simplicity of a simple key press. It already validates that you aren't in an input box, so it will not affect or navigate away from a page when you're in an input box.

PR #1698, opened 2023-12-06
Use FalconPlayer NTP Pool

mjunek:
Remove default Debian pool and use the custom one assigned to the project. Sets this when NTP server is cleared and on initial installation.

### DIFF
--- a/SD/FPP_Install.sh
+++ b/SD/FPP_Install.sh
@@ -462,7 +462,7 @@ case "${OSVER}" in
         ACTUAL_PHPVER="7.4"
         if [ "${OSVER}" == "ubuntu_22.04" -o "${OSVER}" == "linuxmint_21" ]; then
             PHPVER="7.4"
-            echo "FPP - Forceing PHP 7.4"
+            echo "FPP - Forcing PHP 7.4"
             apt install software-properties-common apt-transport-https -y
             add-apt-repository ppa:ondrej/php -y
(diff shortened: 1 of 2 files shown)


dartunder:
I'm just curious why would we need our own pool? What is wrong with the default?

dkulp:
Technically, we are supposed to:   https://www.ntppool.org/en/vendors.html#open-source

SD/FPP_Install.sh
```diff
@@ -1270,6 +1270,15 @@ case "${OSVER}" in
 esac
 
 
+#######################################
+echo "FPP - Configuring NTP Daemon"
+
+# Clear all existing servers and pools and set default pool to be falconplayer NTP Pool
+sed -i '/^server.*/d' /etc/ntp.conf 
+sed -i '/^pool.*/d' /etc/ntp.conf 
+sed -i '\$s/\$/\\npool falconplayer.pool.ntp.org iburst minpoll 8 maxpoll 12 prefer/' /etc/ntp.conf
```
dkulp, line 1279: Since this isn't in a string, should the $ and \ be escaped?  I understand in the settings.php, but it doesn't seem correct here.

mjunek: Yep you're absolutely right, I found & fixed this on my test box but forgot to add the commit!@ Will get it sorted :)

PR #1789, opened 2024-02-03
Added led_row_address_type to FPP for RPi

mortification77:
Added an option not yet implemented by the FPP Interface for the rpi-rgb-led-matrix library called led_row_address_type which allows for usage of additional OUTDOOR LED Panels with Panel-Specific sub-arrangements.


cpinkham:
Can you get rid of the debug line in RGBMatrix.cpp and also change the values in the dropdowns to remove the numbers and parenthesis so it is simpler like the list below?  I don't think users need to see the behind the scenes integer value for the setting.

```
Standard
AB-Addressed Panels
Direct Row Select
ABC-Addressed Panels
ABC Shift + DE Direct
```

mortification77:
Absolutely!

When I was testing this out, I was trying it on my sign panel, which is a Serpentini layout where each row has the panels turned upside down.

I expected the test to show the orientation based on the configuration, but apparently test mode does not handle panel, orientation, and is showing raw orientation.

So basically, I had to remove some other code from my build.

I will make the mod later this evening.

Crazy at work lately!

mortification77:
@cpinkham  I hope I did that correctly!

I code in quite a few languages on a regular basis, but am not great with source control yet.

mortification77:
@cpinkham please let me know if this now looks good.

I am good up to the point of updating my fork, but having to re-merge with the fpp master, that's where I was uncertain!

cpinkham:
Sorry for the delay.  When I viewed the diff in github, it looks like the latest commit may have duplicated some lines in co-ledPanels.php.  Can you check your side to see if they appear to be duplicated?

Here are the lines I'm referring to:
398-401 <-> 402-405
541-543 <-> 544-546
659-660 <-> 661-662
719-720 <-> 721-722

mortification77:
I'm going to redo it in full as I would like to get this completed.

PR #1804, opened 2024-02-18
Healthcheck page updates

OnlineDynamic:
Render on fire fox correctly and fix media directory check

### DIFF
--- a/scripts/healthCheck
+++ b/scripts/healthCheck
@@ -261,13 +261,14 @@ HC_RootDisk() {
 }
 
 HC_MediaDisk() {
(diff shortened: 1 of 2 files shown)


darylc:
So for me, the health check page did OK on Windows Firefox before the change.  After the PR it seems to have a problem.
Before PR:
image
After PR:
image

OnlineDynamic:
@darylc wierdly I get the opposite - without change I get that strange behaviour - I'm on windows 11 firefox 122.0.1 (64-bit)

OnlineDynamic:
@darylc in fire fox if you go to about:config  and search for network.buffer.cache.size what size do you have - mines 32768

PR #1813, opened 2024-02-20
Added RGBMatrix led_row_address_type Option for RPi

mortification77:
This is a REDO of my previous attempt at a PR to add the led_row_address_type Parameter for added LED Panel Compatibility on Raspberry Pi.

### DIFF
--- a/src/channeloutput/RGBMatrix.cpp
+++ b/src/channeloutput/RGBMatrix.cpp
@@ -190,6 +190,22 @@ int RGBMatrixOutput::Init(Json::Value config) {
             options.multiplexing = 0;
         }
     }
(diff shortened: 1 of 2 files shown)


mortification77:
@cpinkham I am not certain what was mangled in the  former PR, so here is a clean one that I have fully tested the feature on.

( As a side task, I plan to learn more about source control :) ).

mortification77:
Previously, I had attempted to not simply push the mods that you and I had discussed, but reset the commit in my repo which made the duplicates MUCH harder to find.

I have now fixed them and it looks much, much cleaner.

mortification77:
Thank you!

PR #1815, opened 2024-03-08
FileManager Updates / Tablesorter upgrade / JQuery Fixes

OnlineDynamic:
This PR does the following

1. Moves the file manager tables into being tablesorter tables to allow sorting of columns
2. Adds column titles
3. adjusts the filesize function to allow correct sorting of columns containing filesizes
4. upgrades tablesorter to v2.32.3
5. seperates inline js to a seperate file for easier coding and debugging
6. renames the file manager page from 'upload.php' to 'filemanager.php'
7. fixes lots of jquery depreciated function calls throughout the UI
8. Displays a no files found message if files of that type don't exist
9. Removes unused docs.php


cpinkham:
Is filemanager.js referenced somewhere?  As I looked through the patch, I saw the new file, but don't see it referenced anywhere.

Also, could you upload a screenshot of what the new UI looks like?  I'm definitely interested in this update, but it would be helpful to see what it does before merging the pull request.

OnlineDynamic:
@cpinkham good spot - put another commit in to get kid of that file I created when testing some stuff

UI looks like:
image
image

cpinkham:
I'm wondering if there is value in optionally allowing enabling the filter boxes.  On systems with a lot of files it might come in handy.

OnlineDynamic:
> I'm wondering if there is value in optionally allowing enabling the filter boxes. On systems with a lot of files it might come in handy. It could be a setting in the UI settings tab and grouped with the Thumbnail size setting.

can do... what's your preference? control in the UI settings tab or directly on screen something like: 
image

cpinkham:
I think a single setting is good enough but it could be shown on the settings page and the file manager to make it easier to turn on/off as needed.

OnlineDynamic:
> Did the Filter box get added yet? I am not seeing it on my system.

Its awaiting  #1820 being merged to turn on filter capability

PR #1820, opened 2024-03-13
Filemanager with filters

OnlineDynamic:
This PR adds filters to the tablesorter tables in filemanager

@cpinkham - as discussed - an additional PR for filtering


cpinkham:
Thanks.

PR #1851, opened 2024-05-25
25 May - Some appearance behaviour improvements

OnlineDynamic:
fixes for sticky header and hopefully Mac display of channel outputs
addition of a fppMinMaxSliderRange css ruleset for Dan's project
minor jquery-ui upgrade
jgrowl upgrade
start a non table based layout for current monitor to make it responsive
fix other outputs tab to show "Select a Type" as default option and prevent adding another new line without configuring first addition


OnlineDynamic:
@dkulp can you please action this PR

darylc (approved):
We should probably merge this then deal with any remaining bugs

PR #1917, opened 2024-08-12
networkconfig.php : escape HTML special chars in scanned SSIDs

zaicik:
Scanned SSIDs with special characters (e.g. "_It's not me it's you_") would break the drop-down HTML. 

Patched using `.replace()` => should work with all browsers (as opposed to `.replaceAll()` which is only supported by recent browsers)

The patch also filters out `undefined` networks which would create an empty <option> entry at the end of the dropdown

### DIFF
--- a/www/networkconfig.php
+++ b/www/networkconfig.php
@@ -177,6 +177,16 @@
             return self.indexOf(value) === index;
         }
 
+        function escapeHtml(unsafe)
+        {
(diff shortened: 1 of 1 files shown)


darylc:
Thanks @zaicik

zaicik:
✌️🫶🎈🎈

PR #2011, opened 2024-10-20
update fpp.js/validateIPaddress() to fail on leading zeroes

jnovack:
Had an issue today (in the Zoom Room) with a user putting in a leading zero to the IP address field because the OLED was showing leading zeroes.

```javascript
validateIPaddress("192.168.1.1"); // true
validateIPaddress("10.0.0.0");   // true
validateIPaddress("01.02.03.04"); // false (leading zeros)
validateIPaddress("256.1.1.1");  // false (out of range)
```

image
image
image

### DIFF
--- a/www/js/fpp.js
+++ b/www/js/fpp.js
@@ -4117,7 +4117,7 @@ function validateIPaddress (id, allowHostnames = false) {
 	var ip = ipb.value;
 
 	var isIpRegex =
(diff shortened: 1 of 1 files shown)


jnovack:
Looks like vscode automatically cleaned up 4578-4590.  Please reject if that's an issue.

ghormann (approved):

PR #2012, opened 2024-10-23
Add fold output processor that reverses a strand in the middle and interleaves the pixels.

jagerkin:
Interleaving pixels makes it easy to inject power from both ends of a strand and easily wire branching paths like icicles.

### DIFF
--- a/src/channeloutput/processors/FoldOutputProcessor.cpp
+++ b/src/channeloutput/processors/FoldOutputProcessor.cpp
@@ -1,0 +1,74 @@
+/*
+ * This file is part of the Falcon Player (FPP) and is Copyright (C)
+ * 2013-2022 by the Falcon Player Developers.
+ *
+ * The Falcon Player (FPP) is free software, and is covered under
+ * multiple Open Source licenses.  Please see the included 'LICENSES'
+ * file for descriptions of what files are covered by each license.
+ *
+ * This source file is covered under the LGPL v2.1 as described in the
+ * included LICENSE.LGPL file.
+ */
+
+#include "fpp-pch.h"
+
+#include "../../log.h"
+
+#include "FoldOutputProcessor.h"
+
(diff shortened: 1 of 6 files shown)


patdelaney:
I'm not entirely sure that I understand the use case, but to make Rick's life easier cand you write something up that we can you in the manual for this feature.

How would we validate/test this enhancement?

jagerkin:
The Fold Output Processor remaps pixels so that a folded in half sequence of pixels appears like it's straight.

folded_icicle_test

I had two main use cases in mind:

1. When you find you need to do power injection in the field, fold over the pixels and plug into both sides from the same end:
```
physically:
         in -> *---*---*---*---*--\
    pwr inj      *---*---*---*---*/

logically:
         in -> *-*-*-*-*-*-*-*-*-*
```

You can handle these in sequencers with varying degrees of difficulty but that's true of most of the output processors. Having another option for handling flickering without having to remodel all your sequences is what I want it for.

---

2. When you need to chain together strands and it would be more convenient to interconnect from the same end, e.g. icicles:
```
normal wiring:
    in -> *   * - *   * - *
          *   *   *   *   *
          *   *   *   *   *
          * - *   * - *   * -> out

folded wiring:
    in -> */ - */ - */ - */ - */ -> out
          |*   |*   |*   |*   |*
          *|   *|   *|   *|   *|
          \*   \*   \*   \*   \*
```

I feel like this would also be real handy for modifying the wiring of complex props that come with prebuilt models, but my use case is entirely the first one.

---

Testing/Validation: I've been testing by folding a string and running a gradient sequence through it, the display testing RGB chase pattern will works fine for testing.

jagerkin:
So, I tried to match the remap UI options (easy enough since I branched from it):
image

But personally, I think Node Count/Node Type instead of Count being channels would be easier to use, so just say the word and I'll switch it.

PR #2173, opened 2025-02-24
Add another USB relay option

bcarriveau:
Add code for LCUS-1 USB relay module, tested on pi working.

### DIFF
--- a/src/channeloutput/USBRelay.h
+++ b/src/channeloutput/USBRelay.h
@@ -8,7 +8,7 @@ typedef enum {
     RELAY_DVC_UNKNOWN = 0,
     RELAY_DVC_BIT,
     RELAY_DVC_ICSTATION,
-    RELAY_DVC_LCUS1
+    RELAY_DVC_CH340
 } RelayDeviceSubType;
 
 class USBRelayOutput : public ChannelOutput {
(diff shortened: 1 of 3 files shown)


bcarriveau:
The LSU1 uses a ch340 usb control chip over the ICstatuon. It's readily asseccable on Amazon. I will in time try multi channel boards with a ch340 and add functionality for multiple channel relay . I was torn to sub name it LSU1 or CH340 because i didn't have a multi channel board to play with.

src/channeloutput/USBRelay.h
```diff
@@ -1,42 +1,31 @@
...
- * included LICENSE.GPL file.
- */
-
-#include <string>
+#ifndef _USBRELAY_H
+#define _USBRELAY_H
 
 #include "ChannelOutput.h"
-#include "SerialChannelOutput.h"
+// Remove #include "SerialChannelOutput.h" since serialutil.h handles serial ops
```
dkulp, line 5: Please restore the use of SerialChannelOutput so that the capes can properly map things.

bcarriveau:
I think I got things back to the way it should be. Sorry, this is all way above my head. This took me entirely too long to figure out 🤣

bcarriveau:
having compile issues not when try to reinstate that into the .cpp. ill close for now and work on it  to clean it up.

PR #2195, opened 2025-03-25
JSON Setting Backup fixes for v8.5

jaredb7:
Collection of additional fixes to the JSON settings backup
Fixes HTTP 500 Error:
- When backing up up "System Settings" 
- If an the "Copy Backups to additional location" setting was used
- In some cases when trying to generate a backup  (added some logging to try understand the real cause, suspect it's probably just a corrupt backup)

Fix Timezone setting not being restored


darylc:
@jaredb7 did you want this merged?

PR #2431, opened 2025-11-02
fix for DPI driven pixels flashing during fppd start and intro of 3D Virt Display

OnlineDynamic:
Fix DPI pixel flash on startup and unwanted triggering

This fix addresses random pixel flashing during fppd startup and config changes by reordering the DPI initialization sequence. Previously, GPIO pins were configured as DPI outputs before the framebuffer contained valid data, causing pins to output garbage that appeared as brief flashes on WS281x pixel strings (which latch and hold their last received color).

Key changes:

- Deferred pin configuration: GPIO pins are now configured as DPI outputs AFTER blank WS281x-encoded data is written to all framebuffer pages, ensuring pins output black pixels from the moment they're enabled
- Smart display enable: Added logic to only enable the display when significant data is detected (10+ non-zero channels for 3 consecutive frames), preventing overlay/status features from triggering unwanted pixel output
- Display lifecycle management: Implemented explicit EnableDisplay()/DisableDisplay() methods in KMSFrameBuffer with state tracking to prevent automatic re-enabling via set_plane fallback during SyncDisplay
- Proper shutdown: Send blank WS281x data before disabling display and reconfigure DPI pins back to GPIO input on shutdown to fully stop output signals
- CRTC disable on init: Explicitly disable CRTC/plane during framebuffer initialization to prevent old data from previous sessions being scanned out

Also fixed pre-existing scope bug in licensing warning code where outputList variable was used outside its declared scope.

Result: Zero unwanted pixel activity during startup, config changes, or idle periods while maintaining full functionality for test mode, playback, and licensed output/latch features.


OnlineDynamic:
closing PR as I have moved the dev of this to the new features repo

PR #2470, opened 2025-12-06
Fix mobile scheduler header to scroll

bcarriveau:


bcarriveau:
Now testing on ipad im not satisfied with its performance,  gonna look back into it.

bcarriveau:
If anyone can test these, that would be wonderful. No rush, the fix works on my end. The devs have been busy and rock!

OnlineDynamic:
Hi - I'm going to close this PR as I have implemented a different fix of adding floatThead jQuery plugin to that table - this keeps it consistent with the rest of fpp - thanks for highlighting the issue though

PR #2504, opened 2025-12-21
Fixes Hard-to-Read Text in FPP

bcarriveau:
*Fixes Hard-to-Read Text in FPP: Changes text color in the top bar and MultiSync rows so it's easier to see on bright backgrounds—goes black on light stuff, stays white on dark. 

*Doesn't mess with main FPP looks.
Turn It On/Off Easily: Adds a checkbox in Settings > UI > UI Colors to enable it (on by default); JS code only loads if you checked

*Top Bar Tweaks: In menuHead.inc, fixes text and flips logos to black on bright backgrounds; skips the orange update part so it stays normal.

*MultiSync Table Stuff: In multisync.php, watches the table rows and auto-fixes them when it refreshes, keeps things readable without constant checks.

**Note: Have not found an elegant solution for the Download Available icon. Orange on Orange/Dark Orange/Coral are the 3 that are hard to read but i like the OG orange.  Open to suggestions or help with that. I played with the idea of having it get a shadow outline.

### DIFF
--- a/www/multisync.php
+++ b/www/multisync.php
@@ -2797,15 +2797,17 @@ function hexToRgb(hex) {
     const num = parseInt(hex, 16);
     return { r: num >> 16, g: (num >> 8) & 255, b: num & 255 };
 }
-
 function getLuminance(rgb) {
     return (0.299 * rgb.r + 0.587 * rgb.g + 0.114 * rgb.b) / 255;
 }
-
 function adjustMultiSyncRows() {
     const rows = document.querySelectorAll('#fppSystemsTable tr.systemRow');
     rows.forEach(row => {
         let bgColor = window.getComputedStyle(row).backgroundColor;
+        // Skip rows with no background (transparent or rgba with alpha 0) - leave defaults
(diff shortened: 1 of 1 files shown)


OnlineDynamic:
I like the idea of improving readability but think this would  be done better with propper css rules rather than inline javascript manipulation.. personally would be more likely to merge that approach

bcarriveau:
Originally, my idea was to make it selectable (on/off) to get feedback without pushing it on everyone. If there was an issue, they could just disable it, and if the feedback was more positive or there was no issue, then change it over permanently with standard CSS. I can work on just a css update instead no rush!

OnlineDynamic:
personally I find the white on orange easier to read than black, but agree the white on green is hard to read

bcarriveau:
> personally I find the white on orange easier to read than black, but agree the white on green is hard to read

I can make some regular CSS style changes in the coming days and get some feedback again. I'll try to get it in for 10.x. I'm colorblind, so I'm already the odd one out lol.

Issue #2459, opened 2025-11-29
Hang, possible memory corruption, while show is playing


jkriegshauser:
**NOTE:** IF YOU DO NOT ATTACH YOUR LOG FILES, THEN THERE IS A HIGHER PROBABILITY THE ISSUE WON'T BE RESOLVED.  (Steps at end) 
# Please answer the following:
**FPP Version and Hardware**: 
FPP v9.3

**Describe the bug and Steps to reproduce**: 
A clear and concise description of what the bug is and the steps to be taken to reproduce. 
1. Restarted fppd after a crash at 5:32 pm
2. Restarted fppd after a settings upload at 5:41 pm
3. At 7:55 pm, see a bunch of messages in the log about Audio has stalled. Show appeared to be hung.
4. Restarted fppd (just the process, not the device) and show started over playing fine.

**Expected behavior**: 
It shouldn't hang

**Additional context**: 
In the syslog I see: 
`2025-11-28T19:55:10.330735-08:00 FPP-Porch fppd[25355]: malloc_consolidate(): unaligned fastbin chunk detected`
Memory corruption?
I attached gdb before restarting fppd and recorded the fpp_callstacks.txt below. Concerning:
```
Thread 12 (Thread 0xa62029e0 (LWP 28984) "MHD-worker"):
#0  __libc_do_syscall () at ../sysdeps/unix/sysv/linux/arm/libc-do-syscall.S:46
#1  0xb66387ae in futex_wait (private=0, expected=2, futex_word=0xacc00018) at ../sysdeps/nptl/futex-internal.h:146
#2  __GI___lll_lock_wait_private (futex=futex@entry=0xacc00018) at lowlevellock.c:34
#3  0xb6647cc2 in __GI___libc_malloc (bytes=bytes@entry=4096) at malloc.c:3321
#4  0xb662c8f2 in __GI__IO_file_doallocate (fp=0xacc28a50) at filedoalloc.c:101
#5  0xb66365dc in __GI__IO_doallocbuf (fp=0xacc28a50) at libioP.h:947
#6  __GI__IO_doallocbuf (fp=fp@entry=0xacc28a50) at genops.c:342
#7  0xb6635dd0 in _IO_new_file_overflow (f=0xacc28a50, ch=-1) at fileops.c:744
#8  0xb66353d6 in _IO_new_file_xsputn (n=92, data=<optimized out>, f=0xacc28a50) at libioP.h:947
[...]

PR #2509, opened 2025-12-24
Fix crashes in SDLOut (#2459)

jkriegshauser:
Fixes crashes due to data races on SDL::data and handles error results from swr_convert

Also:
* Fixes for building docker-compose-dev on Ubuntu
* Crash timeout was too short for BBB
* Adds tsan/asan build variants


jkriegshauser:
> Can you please check and test?

Seems good to me on a quick test.

> In the future, don't combine a bunch of things into a single PR as it makes it very difficult to review and apply.

Will keep this in mind.

Thanks!

PR #2510, opened 2025-12-24
Allow installation of incompatible plugins

bcarriveau:
Hide "Allow installation of incompatible plugins" checkbox in Basic & Advanced UI modes

Only show the checkbox then Experimental or Developer mode

This allows "risky" installs :-)

Note: The existing compatibility detection logic treats plugins with maxFPPVersion="0" (or empty/no upper bound) more permissively than those with an explicit upper bound (e.g. "9.99"). This can lead to a counter-intuitive situation where very old plugins with no declared upper limit appear "safer" or more compatible than plugins that explicitly state their last tested version.

A future enhancement could consider treating maxFPPVersion="0" more conservatively (e.g. as "incompatible" after a certain number of major versions) to better reflect real-world risk and encourage plugin maintainers to update their pluginInfo.json files and test against current FPP releases.
2025-12-24 12_23_51-FPP - Test
2025-12-24 12_23_30-FPP - Test

### DIFF
--- a/www/plugins.php
+++ b/www/plugins.php
@@ -1,35 +1,28 @@
 <!DOCTYPE html>
 <html lang="en">
-
 <head>
     <?php
     include 'common/htmlMeta.inc';
     require_once('config.php');
     require_once('common.php');
-
     writeFPPVersionJavascriptFunctions();
-
     include 'common/menuHead.inc';
     ?>
     <script>
         var installedPlugins = [];
         var pluginInfos = [];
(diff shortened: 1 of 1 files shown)


OnlineDynamic:
Not sure we would want to do this - if a plugin is explicitly not compatible with current version as declared by it's developer then what advantage does this give? - If a plugin needs fixing a developer would know how to manually install it without the UI and would be able to contribute to its source repository to get it into a state where it would be compatible

PR #2515, opened 2025-12-26
fix command scheduler showing null and javascript error from undefined

agent462:
## Test Core Problem
In scheduler.php:
To replicate before fix:
* Add row in Scheduler
* Select Schedule Type -> Command
* Save
* Refresh the page, a js error exists and you will now have "null unavailable" in your dropdown list.

After fix:
* Add row in Scheduler
* Select Schedule Type -> Command
* Save
* Refresh the page, NO js error exists and you will now have NOT have "null unavailable" in your dropdown list.

## Also fixes a global scoped variable (missing var).
```
// Before
e = {};
// After
var e = {};
```

### DIFF
--- a/www/scheduler.php
+++ b/www/scheduler.php
@@ -498,7 +498,7 @@ error_reporting(E_ALL);
 
         function GetScheduleEntryRowData(item) {
             var schType = $(item).find('.schType').val();
-            e = {};
+            var e = {};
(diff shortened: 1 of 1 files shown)


OnlineDynamic:
Have tested - there is no js error in current master and it is correctly telling the user a null (blank) command has been saved and highlighting the dropdown in red to show the error state so don't think this PR is needed

Issue #2277, opened 2025-07-06
Playlist Sort Options


kevinsaucier:
I searched and I'm kind of amazed I couldn't find anything so I'm almost thinking I'm just stupid and don't know how to do it.  If that's the case, please point it out.  If not, can we please please please have a few more sorting options for Playlists?

At the very least, I'd like to be able to Sort Alphabetically.   This is incredibly useful for Remote Falcon playlists.  It might also be nice to sort by Size, Date Added, and maybe Last Modified as well.  Also, a Date Added and/or Last Modified column would be a helpful addition to the columns, potentially with a header bar instead of labels inline.

Thanks!

<img width="440" height="418" alt="Image" src="https://github.com/user-attachments/assets/56336e0b-49e0-4219-8122-1da8e9910196" />

Poporacer:
I am not sure why you would want a **Playlist** to be sorted alphabetically or in any of the other methods you requested? I do not know of anyone that would play their show in an alphabetical order or in the order it was modified? 
In regard to Remote Falcon, you can display your playlist in alphabetical order, Remote Falcon has that functionality built in.
If you go to File Manager you can see your Sequences in just about any sort order that you want.

kevinsaucier:
For the RF playlist, it's difficult to keep track of what tracks have been added to the playlist already without being able to sort it.  This leads to duplicates and/or missing tracks.  Alpha sort on the RF side is helpful for the RF interface, but not for managing the playlist itself.  Same with being able to sort tracks by date.....it's much easier to see if a track is older than it should be by sorting by date and then looking at the top/bottom.  Doing either of these things in the file manager is not really feasible.  Since we already have the 'Randomize' option in the drop down (which, personally, I can't see why anyone would want to use), seems like it would be pretty easy to add at least the alpha sort without having to make any other major changes.

kevinsaucier:
K, so I'm not a Javascript programmer by any means, but I was able to copy/cobble the below and it works.  I poked around and I don't think I'm missing anything but most of this language is foreign to me.  😬  I'm also not a Git person so I'm not quite sure what's involved in forking/contributing.  Hoping this will at least spark someone to include/proofread the code so I don't have to edit every time I upgrade.  😁

<img width="582" height="832" alt="Image" src="https://github.com/user-attachments/assets/79aad894-f7de-4c5b-8898-a76fc0c8b448" />


In /js/fpp.js:  (I put it right above function RandomizePlaylistEntries ())

`
[...]

PR #2522, opened 2025-12-28
Playlist sorting for ASC and DESC

agent462:
#2277

I find this useful and have been manually doing this for my RF playlist.  User that submitted the enhancement request explains the downsides of trying to do this exclusively in RF.

Note: There is an existing issue with the playlists page in master that does not allow the buttons to be interactive for up to several seconds in multiple browsers that I've noticed with large playlists.  I've poked at the issue but haven't found a fix.  I will create a separate issue for this but would also love your opinion and whether you've looked at this @OnlineDynamic so I don't waste time.

## Enhancement Added
This will ASC or DESC sort Lead In, Main Playlist and Lead Out lists independently for all entry types.  

I have changed the randomizeBuffer to playlistBuffer so it's reusable and follows the same pattern and not needing a secondary span for another buffer.

## Testing

1. Load up a playlist with various entries including a subsection like Lead In
2. Click Playlist Actions -> Sort Playlist ASC | DESC
3. Playlist is sorted correctly
4. Save Playlist
5. Reload Playlist

### DIFF
--- a/src/commands/Commands.h
+++ b/src/commands/Commands.h
@@ -15,6 +15,7 @@
 #include <httpserver.hpp>
 #include <list>
 #include <map>
+#include <span>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


OnlineDynamic:
As per Rick's comment in the original ticket - I'm still not seeing why you would ever want to sort a playlist Alphabetically, hence why this hasn't been implemented previously

OnlineDynamic:
have closed the request

Poporacer:
@kevinsaucier You are correct that we welcome PR requests, but if the developers were to implement every PR request that was submitted, then the FPP code would end up being huge and full of bloat. Decisions need to be made on whether it helps the masses or the few. Or if it is easily accomplished via existing methods. It also has to be compliant on mobile devices. I wasn't involved in the decision, but as I mentioned before, I highly doubt there are that many people that actually play their playlists in an alphabetical order.

PR #2523, opened 2025-12-28
Fix some memory leaks

jkriegshauser:
* Fixed: Historically, the GPIO command was leaked because of a crash, though this is no longer happening in local testing. If it recurs, we should debug the crash (which may be easier with ASan support now)
* Fixed: Leaking dict (and possible leaking channel layouts) in SDL code
* Fixed: ScheduledItem leaks in the Scheduler
* Changed: ScheduledItem and vectors in the Scheduler are no longer separate heap allocations, which should make management easier, reduce heap usage and fragmentation, etc.

### DIFF
--- a/src/Events.cpp
+++ b/src/Events.cpp
@@ -13,11 +13,13 @@
 #include "fpp-pch.h"
 
 #include <algorithm>
+#include <chrono>
 #include <climits>
 #include <ctime>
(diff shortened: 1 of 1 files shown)


jkriegshauser:
@dkulp thanks! I don't mind making tweaks too if different things are preferred :D

PR #2527, opened 2025-12-30
fix large playlists overflowing into the wrong section (lead out), fix stale DOM data from being put in new playlists

agent462:
Large playlists were overflowing into Lead Out and counting was not correct with large playlists.

Batch processing used var in a loop with setTimeout, causing JavaScript closure issues where all async callbacks referenced the final loop values instead of their intended values

Also fixed the main playlistInfo.total_items to count all items not just main.

Screenshot 2025-12-29 at 9 00 16 PM

### DIFF
--- a/www/js/fpp.js
+++ b/www/js/fpp.js
@@ -6487,23 +6487,17 @@ function PopulatePlaylistDetails (data, editMode, name = '') {
 
 				let batchIndex = 0;
 				let startingEntryCount = entries; // Capture entry count before async processing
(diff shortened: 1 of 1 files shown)


OnlineDynamic:
I need to check with the xlights guys on the playlist total count.. i left it counting just the main playlist items as i think that is the logic being used by xlights when it pushes sequences into a playlist... agree doesnt seem correct but think thats because xlights doesn't think about leadin and outs... the DOM thing seems to be very browser specific.. as most browsers i'm not seeing it... only managed to reproduce by running chormium

PR #2541, opened 2026-01-12
Add USB mount button with dropdown/verification and unmount refresh; fix dialog CSS typo

bcarriveau:
- Added a "Mount USB" button with dropdown to select unmounted devices (using GetAvailableBackupsDevices), including dialog confirmation for safety.
- Added page refresh to force unmount success case to reflect UI changes consistently (matches mount behavior).
- Fixed CSS typo in dialog-confirm elements: "flat:left;" -> "float:left;" for proper icon alignment.
2026-01-11 19_36_58-NVIDIA GeForce Overlay

### DIFF
--- a/www/api/controllers/files.php
+++ b/www/api/controllers/files.php
@@ -14,7 +14,7 @@ function MapExtention($filename)
         return GetDirSetting("images");
     } else if (preg_match("/\.(eseq)$/i", $filename)) {
         return GetDirSetting("effects");
-    } else if (preg_match("/\.(sh|pl|pm|php|py)$/i", $filename)) {
+    } else if (preg_match("/\.(sh|pl|pm|php|py|js)$/i", $filename)) {
         return GetDirSetting("scripts");
     } else {
         $pluginDirectory = GetDirSetting("plugins");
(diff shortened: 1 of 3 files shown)


jessica12ryan:
My take on this: So if the ability to mount drives is available, we should add the ability to unmount (without force)

This is what I was thinking:
1) Remove the warning "Use caution when unmounting USB devices as they may be in use by the system processes. Forcing a unmount may result incomplete data or data loss."

2) Change the "Force Unmount" button to "Unmount"

3) Create a popup dialog that says:

    "You are about to unmount a disk...
    
    Use caution when unmounting USB devices as there may be in use by the system processes. Forcing an unmount may result in data loss."
    
    With 3 buttons "Force Unmount" probably in red justified to the left, and two other buttons justified to the right - "Cancel"  and "Unmount"

I think this would improve the new change while providing a more stable way to unmount drives. And by moving that yellow warning into a dialog, it will help make the page look cleaner
updated

bcarriveau:
@jessica12ryan , something like this could work?

2026-01-16 17_45_37-FPP - Masturator2

bcarriveau:
While trying to diagnose why FPP backup file copy was ignoring a config file from a plugin, I found that FPP offers a way to force unmount a device, but no way to mount a USB drive without going through a reboot after changing the main device storage just to view its contents. The use case is probably minimal for most, but it might tie or exceed the force unmount.

Now that I think about it even more, it would make even more sense to have it mount to the /home/fpp/media location, as the way it currently would run would be a bit in the back 40 acreage. It would also allow it to show up with a simple browser FTP login. Even better if a tab in the file manager appeared when mounted to view it all right from the interface already open.

patdelaney:
another possible use case although maybe not that common, is to pull files that may be on an sd card that maybe will not boot that was in a FPP.

PR #2580, opened 2026-02-24
api/schedule: add generic opt-in post-save hook callback

robbiet76:
This patch adds a generic, opt-in, localhost-restricted POST JSON hook in schedule.php and wires it into SaveSchedule().

### DIFF
--- a/www/api/controllers/schedule.php
+++ b/www/api/controllers/schedule.php
@@ -38,11 +38,112 @@ function SaveSchedule() {
 
         //Trigger a JSON Configuration Backup
         GenerateBackupViaAPI('Schedule was modified.');
+        try {
+            TriggerScheduleSaveHook();
+        } catch (\Throwable $e) {
+            error_log('[schedule] post-save hook threw: ' . $e->getMessage());
+        }
 
         return GetSchedule();
     }
 
     halt(404, 'Unable to open schedule.json for writing.');
+}
+
+/////////////////////////////////////////////////////////////////////////////
+// Optional post-save hook. Never blocks/fails schedule save.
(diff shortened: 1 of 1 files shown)


darylc:
@robbiet76 Sorry about taking to long to reply.

Rather than adding a new single-purpose curl-based hook here, what about triggering an existing FPP command preset instead — e.g.   CommandManager::INSTANCE.TriggerPreset("SCHEDULE_SAVED"), the same pattern already used for FPPD_STARTED/FPPD_STOPPED (fppd.cpp),  SEQUENCE_STARTED/SEQUENCE_STOPPED (Sequence.cpp), and OUTPUTS_ENABLED/OUTPUTS_DISABLED (OutputMonitor.cpp)?

A few advantages over the ScheduleSaveHookUrl setting approach:
  - Multi-consumer by default — a preset is a named list of commands, so any number of independent integrations can add their own command to  SCHEDULE_SAVED without clobbering each other. A single URL setting only supports one consumer at a time.
  - No new config surface — users already configure preset actions through the existing Presets UI, so there's no need to hand-edit settings.json/hit the settings API to set an undocumented key.
  - The HTTP use case isn't lost — if someone actually wants to POST to a local listener on schedule save, they just add an HTTP Command to the SCHEDULE_SAVED preset. It becomes one option among many rather than the only option.
  - Much smaller diff — this becomes roughly a one-line call instead of ~100 lines implementing URL validation, curl, timeouts, and error-log  plumbing that duplicate what Commands CommandManager already provide.

Feel free to send a new PR along those lines and we'll review it promptly.

PR #2584, opened 2026-03-05
Using MDNS/avahi register FPPD as a service and discover other FPPD s…

NathanKulp:
…ervices on the network


dkulp:
This was merged.

PR #2594, opened 2026-04-03
added servo api command

NathanKulp:
created servo api command


darylc:
@dkulp what shall we do with this one?

PR #2608, opened 2026-04-24
Refactor CI to use native actions and update versions

jnovack:
* Ensures `latest` does not run on feature branches
* Uses steps output rather than environment to support multi-job builds in the future
* Uses native official github actions
  - Setting up QEMU
  - Creating Image Tags
  - Creating Labels
* Update all actions to avoid Node 20 deprecation

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -4,13 +4,13 @@ on:
   # Run when pushes to branch or create new Tag
   push:
     branches:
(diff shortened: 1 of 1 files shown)


jnovack:
@dkulp Want to make sure this is actually what you intended.

Big fix here is that a push to a feature branch or tag (like 9.1 or 9.5) does not update `:latest`.  Only `master` updates `:latest`

PR #2616, opened 2026-04-29
Fix (nearly) all scss warnings

jnovack:
A (now fewer ;)) significant deprecation warnings were occurring with the SASS build for the next version release. This updates all commands.

Importantly **NO MODIFICATION WAS DONE TO `fpp-bootstrap-5-3.css`** (only to the map file).  This results in ZERO generational changes.

A significant refactor will need to be done in the future to change the `@import` statements, that is beyond this scope, as it only affects Dart Sass v3.0.0 (official release is still on the v1 branch).


jnovack:
What changed:

* local darken() / lighten() calls moved to color.adjust(...)
* local mix() calls moved to color.mix(...)
* the two local if($enable-gradients, none, null) property cases were rewritten to explicit @if
* _spacing.scss got the minimal local map.keys(...) namespacing change

PR #2627, opened 2026-05-03
MultiSync: honor secondsElapsed in media START packets

### LINKED ISSUE #2626: MultiSync media START packets ignore secondsElapsed (remotes always start mid-stream playback at frame 0)
## Summary

When a remote receives a MultiSync media `START` packet with non-zero `secondsElapsed`, the value is dropped — the remote always starts media playback at frame 0. The first subsequent `SYNC` packet then triggers `libvlc_media_player_set_time` (the broken seek path noted at `VLCOut.cpp::AdjustSpeed` — *"jumping in vlc is problematic and causes video freeze"*), producing a visible 1–2s jump (on libvlc 3.x) or a decoder freeze (on the libvlc 4 dev build present in some FPP images).

## Where the value is dropped

`src/MultiSync.cpp` `ProcessSyncPacket()` for `SYNC_FILE_MEDIA` / `SYNC_PKT_START`:

```cpp
case SYNC_PKT_START:
    StartSyncedMedia(spkt->filename);  // spkt->secondsElapsed is never read
```

The receive-side infrastructure to honor a start position **already exists** in v9.x:

- `MediaOutputBase::Start(int msTime)` accepts the offset
- `VLCInternalData::startPos` + `seekableEventCallBack` issue the seek when libvlc enters the Seekable state (the seek path that fires before the first decoded frame and works reliably)
- `SDLOutput::Start(int msTime)` honors msTime via buffer skip + video-frame discard

Only the plumbing from MultiSync is missing.

## Reproduction

Send a `SYNC_PKT_START` for `SYNC_FILE_MEDIA` with `secondsElapsed=30.0`. The remote plays from frame 0 instead of from 30s. With debug Sync logging enabled you can see the value arrive at `ProcessSyncPacket` and never get passed to `StartSyncedMedia`.

## Tested fix

Reproduced and validated on Pi 4 + Falcon Player image v2025-11 (FPP 9.5.2, libvlc 3.0.21), hardware decoding ON. With the patch, mid-video START arrives at the requested position immediately, no jump. Multiple 1080p H.264 files (.mov/.mp4, varying length and bitrate) all confirmed working.

PR with the fix to follow.

## Related

#883 (closed) — same `set_time` "video freezes / audio continues" symptom but reported via SYNC drift correction. Initializing the player at the correct position from START avoids reaching the broken path on the seek-on-start case.

djkoren:
Fixes #2626.

## Summary

Plumbs `secondsElapsed` from MultiSync's media `START` dispatch through `StartSyncedMedia` → `StartMediaOutput` → `mediaOutput->Start(msTime)`. The receive infrastructure (`VLCInternalData::startPos`, `seekableEventCallBack`, `SDLOutput::Start(int)`) already existed; only the plumbing from MultiSync was missing.

Also forwards the value in `UpdateMasterMediaPosition`'s file-change branch so a master switching to a different media file mid-playback gets the same correct seek-on-start treatment. `m_remoteOffset` is applied to mirror the existing handling in `SYNC_PKT_SYNC`.

4 files, +14 / -9. Source-compatible (default-valued parameter); no plugin headers affected.

## Test plan

- [x] Regression: play from beginning unchanged
- [x] Mid-video START on three 1080p H.264 files (24/30 fps, .mov/.mp4) — all start at the requested position with no visible jump
- [x] Drift correction continues to work after seek-on-start

Tested on Pi 4, Falcon Player image v2025-11 (FPP 9.5.2, libvlc 3.0.21), hardware decoding ON.

### DIFF
--- a/src/MultiSync.cpp
+++ b/src/MultiSync.cpp
@@ -2445,12 +2445,12 @@ void MultiSync::OpenSyncedMedia(const std::string& filename) {
     OpenMediaOutput(filename);
 }
 
-void MultiSync::StartSyncedMedia(const std::string& filename) {
-    LogDebug(VB_SYNC, "StartSyncedMedia(%s)\n", filename.c_str());
+void MultiSync::StartSyncedMedia(const std::string& filename, float secondsElapsed) {
+    LogDebug(VB_SYNC, "StartSyncedMedia(%s, %.2f)\n", filename.c_str(), secondsElapsed);
     for (auto a : m_plugins) {
         a->ReceivedMediaSyncStartPacket(filename);
     }
-    StartMediaOutput(filename);
+    StartMediaOutput(filename, secondsElapsed);
 }
 
 /*
@@ -2536,7 +2536,11 @@ void MultiSync::ProcessSyncPacket(ControlPkt* pkt, int len, MultiSyncStats* stat
             stats->pktSyncMedOpen++;
             break;
         case SYNC_PKT_START:
-            StartSyncedMedia(spkt->filename);
+            secondsElapsed = spkt->secondsElapsed - m_remoteOffset;
+            if (secondsElapsed < 0)
+                secondsElapsed = 0.0;
+
+            StartSyncedMedia(spkt->filename, secondsElapsed);
             stats->pktSyncMedStart++;
             break;
         case SYNC_PKT_STOP:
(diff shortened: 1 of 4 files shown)


OnlineDynamic:
we are about to remove vlc and some of sdl as part of the move to gstreamer and pipewire so I suggest we hold fire on this PR until that change comes across into master ffrom the new-features repo... might need reworking post that

djkoren:
Sounds good! I'll re-test when that update drops.

darylc:
@djkoren FYI the update Online Dynamic mentioned is now merged into master.

rajumsrana16-byte (approved):

darylc:
See comments in #2626

Issue #2613, opened 2026-04-27
fpp 10 about/upgrade page recommendation


bloojoop:
on FPP build 10.x-master-213-ge59ef994, if both a OS and a FPP upgrade are available, there is a recommendation that doesn't seem to read correctly.  

<img width="3094" height="638" alt="Image" src="https://github.com/user-attachments/assets/6570b7f5-c883-4b82-8675-5f395d30c717" />

Is this recommendation accurate ??  both an fpp and an 'OS' update is available, shouldn't we actually do a OS update first then if necessary the FPP update.  Since the OS update contains the version of FPP that was available at the time the OS was packaged.  This recommended path will possibly leave the customer on an older version of fpp... example fpp 10.1.3 (made up version) is available but the OS only has 10.1, if doing fpp first, then it will be 10.1.3 and then if proceeding to do the OS update, fpp will only be 10.1 (an older version).

darylc:
@agent462 we're going to consolidate feedback about the about page into this issue, I can't seem to assign it to you though, can you try to self assign it?

patdelaney:
Got some feedback about the About Page.  When in Advance Mode please move the Version information to the top Above Update FPP software.

<img width="1348" height="791" alt="Image" src="https://github.com/user-attachments/assets/522e7f4a-cccc-436b-9614-51734369b209" />

patdelaney:
@darylc   Yes, you can assign tasks (issues or pull requests) to a user who is not a project member, but only if they have commented on the issue or PR first.

patdelaney:
@agent462  Please comment on this issue so we can assign it to you

agent462:
I can look at this - edited my comment because this only shows in advanced.

agent462:
I need to rebuild my dev box so this will likely be this weekend and I'll have a PR for it.

OnlineDynamic:
@agent462 might be worth also getting together with @jnovack who has been also looking at improving the css handling (and adding in dark mode) and he has some thoughts on patterns and anti-patterns and is also in process of implementing playwright from a UI testing prespective

bloojoop:
another note about this page,, when selecting the pi nightly fppos from the drop down and selecting on the release notes, the text is unreadable. 

<img width="1626" height="891" alt="Image" src="https://github.com/user-attachments/assets/b7ea32c7-d51e-4565-bb23-d37622a29eec" />


[...]

PR #2637, opened 2026-05-09
System-Upgrade.php tweaks - Tighten system upgrade page layout and improve advanced UI

agent462:
- Tighten padding/margins on cards, banners, and info boxes for a more compact vertical layout on the system upgrade page; move the OS reboot warning into the "When to use" box
- Collapse the "When to use / What it does" info grids into <details> elements when uiLevel >= 1 to reduce clutter for advanced users (basic UI still shows them expanded inline)
- Move the Version Information card out of the Advanced section so it's always visible at the top of the page, and reposition the donate banner below the upgrade options
- Fix release notes modal styling and handle nightly builds correctly

### DIFF
--- a/www/css/fpp-system-design.css
+++ b/www/css/fpp-system-design.css
@@ -1321,6 +1321,58 @@
 	line-height: 1.5;
 }
 
+/* Collapsible wrapper for info-grid (advanced UI mode) */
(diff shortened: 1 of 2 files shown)


agent462:
#2613

PR #2638, opened 2026-05-10
FPP_Install.sh - Implement seamless migration to systemd-networkd, plus failsafe tweaks

jessica12ryan:
To fix issues with FPP_Install.sh:
Added logic to seamlessly migrate from NetworkManager to systemd-networkd without dropping connectivity. Enhanced network configuration handling and ensured required packages are installed.

When installing on a base version of Debian 13 Trixie using the "--img" flag, the install fails as systemd-networkd is disabled by default and NetworkManager is in use. These changes will fix this by examining which service is in use, and cleanly migrating to systemd-networkd if necessary without losing connection. Previously, the script would call commands to systemd-networkd with no regards to whether or not that was the running network daemon. If it was not, the script would fail. This is a clean fix to move everything to systemd-networkd if NetworkManager is in use.

Also added checks when disabling display managers to check if the display managers are installed before attempting to disable - preventing errors from base installs where no GUI is present.

Also ensures curl is installed before executing curl commands as it is not packaged with base Debian 13 and makes sure that ssh is installed. With these changes, the script will now install on the most minimal install of Debian 13 with the --img flag.

And last but not least, a tweak to capture the install script name and install script directory. Previously, the script made references to /root/FPP_Install.sh which would leave an error if not installed from that directory. You can now install the script from any directory or rename the script and it will still install.

### DIFF
--- a/SD/FPP_Install.sh
+++ b/SD/FPP_Install.sh
@@ -726,9 +726,10 @@ install_base_packages() {
             # generic display-manager alias cover the desktop-DM cases we
             # might encounter when --img is run on a desktop install. The
             # legacy gdm/wdm/xdm names are gone from supported releases.
-            systemctl disable gdm3
-            systemctl disable lightdm
-            systemctl disable display-manager.service
+			# Check if exists before disabling using cat
+			systemctl cat gdm3 >/dev/null 2>&1 && systemctl disable gdm3
+			systemctl cat lightdm >/dev/null 2>&1 && systemctl disable lightdm
+			systemctl cat display-manager.service >/dev/null 2>&1 && systemctl disable display-manager.service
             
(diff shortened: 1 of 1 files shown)


SD/FPP_Install.sh
dkulp, line 1: Can I ask what the use case for this is?    The install instructions are specific to download and run the FPP_Install.sh script and it produces an FPP_Install.log.

dkulp:
Can I ask what the use case is for this?   This seems to just add extra variables for no reason.   All the install docs and such specifically state FPP_Install.sh and the log is that so there doesn't seem to be any reason to complicate the script with these changes.

jessica12ryan:
It was just to make the installer more universal. There is no need to make those changes if you feel that it adds too much complication.

PR #2641, opened 2026-05-12
feat: add scalar for API documentation

jnovack:
Endpoint - `<host>/api`

DEPRECATES:
- `apihelp.php`
- `help.php`
- `endpoints.json`

Developers are expected to write phpDoc (docblock) style comments.  See `www/api/README.md` for more info.

PHASE 1 is DOCUMENTATION ONLY.  No API changes were made.

**Random Endpoint**
image

**Test Request**
image


OnlineDynamic (commented):
@jnovack will pull this down later to test it out... couple of thoughts... Will check when I have it down locally but a) is removing endpoints.json safe? Does anything else reference that and b) is it possible to fpp theme the scalar pages? As your screen shot of the api docs doesn't look very 'fpp', looks out of place currently

OnlineDynamic:
In fact the more I think about it having the standard header and nav would be useful to properly incorporate it

OnlineDynamic:
@jnovack - another thought - how do plugins register their api's? Would be good to standardise so the documentation page could show the standard fpp api's as well as those provided by the installed plugins

jnovack:
@OnlineDynamic 

> Anything referencing `endpoints.json`?

Documentation only.

> Possible to theme

Yes.  What are our colors?  (Also replied in Slack)

> ...standard header and nav...

Whoa, how deep are we going?  All the way down to "pageContent"?

> how do plugins register their api?

`README.md` has instructions on how to write phpDoc compliant headers to the php pages.

PR #2642, opened 2026-05-14
Round 1 - API VERB Inconsistencies

jnovack:
HTTP `GET` is defined as a safe, idempotent method. The routes listed in this section all mutate system state — they start or stop playback, reboot hardware, run scripts, trigger events, or modify files — and were therefore incorrectly exposed as `GET`. All have been moved to `POST` (or, in one case, corrected from `POST` to `GET`).

Unless noted otherwise, the only required client change is updating the HTTP method. Parameters that were previously passed as URL query strings remain in the query string. No request body is required.

This introduces a `MIGRATION.md` file to keep track of the changes to v10, and starts to organize the helper classes.

More rounds will follow, I figured it would be good start low and slow so I can course correct now rather than massively rewrites later.  Additionally, some functions are slated for multiple changes, and I would rather than do them in logical batches than complete rewrites.

Acceptance (merging) here and I'll start on the next logical batch based off this.


OnlineDynamic (changes requested):
As discuss on slack, this needs the version ingredients approach implemented before we can merge in

Issue #2648, opened 2026-05-17
OS Upgrade Confusion.


ghormann:
On a  Raspberry Pi running v2026-02  with v10.x-master-495-g1a6f7820 (master branch).    The New system-upgrade page appears confused.  The banner states that there is an upgrade available, yet further down the page it shows the OS is up to date. 

<img width="1965" height="861" alt="Image" src="https://github.com/user-attachments/assets/0ff71539-a2a7-425d-884f-74971af87e85" />

jessica12ryan:
On the Pi, you are using a development version of FPP (10.x-master), so there is actually a newer OS in the list which is Pi-nightly_2026-05.fppos. I'm guessing because you are using a dev branch, you'll get dev OS updates too, same as you get dev FPP updates. It just seems like the banner is reflecting this while the badge is not.

darylc:
The check needs to ignore nightly

darylc:
@agent462  can you comment so I can assign please?

ghormann:
1) I agree that nightly should be ignored
2) The fact that we have two different methods of computing this feels wrong and something that should be in a function or API call.

agent462:
here

cybercop23:
We should have consistency between buttons and pill colors. If *any* updates, button should be yellow, if no updates, grayed out.
OS Version pill in Upgrade OS should be green, to match the the upper banner, or both gray.

I saw it before, if both an upgrade and an OS, the path should be OS -> upgrade (if anything), no need to upgrade FPP first.

darylc:
@agent462 Can you have a look at the colors on the page - as @cybercop23 said, it is a bit confusing to have 'Up to date' and 'Update available' both in Green right near each other.  Can you have a think about this and see what alignment can be done please?

<img width="901" height="854" alt="Image" src="https://github.com/user-attachments/assets/608d9d60-7d32-4ab6-b701-26213f263606" />

darylc:
The issue that this was created about is now resolved.  The color issues are also being discussed in https://github.com/FalconChristmas/fpp/issues/2613 so we will continue there and close this one out.

patdelaney:
@jnovack please keep an eye on this as well because this is related to other color issues in multisync page and what ever we choose should be the same across the board.

PR #2653, opened 2026-05-20
System upgrade updates based on feedback

agent462:
Fixes for feedback in #2613 and #2648

- sync OS version badges with detected state
- skip nightly OS
- align upgrade buttons and pills with update state
- change upgrade recommendation logic when both OS and FPP exist

For testing:
http://<your-fpp>/system-upgrade.php?test=uptodate
http://<your-fpp>/system-upgrade.php?test=commit
http://<your-fpp>/system-upgrade.php?test=branch
http://<your-fpp>/system-upgrade.php?test=both
http://<your-fpp>/system-upgrade.php?test=osonly
http://<your-fpp>/system-upgrade.php?test=major
http://<your-fpp>/system-upgrade.php?test=eol

### DIFF
--- a/www/api/controllers/system.php
+++ b/www/api/controllers/system.php
@@ -276,7 +276,8 @@ function GetUpdateStatus()
                 "latestMajorVersion" => $currentMajor
             ));
         } else if ($test === 'both') {
-            // Simulate: branch upgrade AND commits behind
+            // Simulate: FPP branch upgrade AND OS upgrade both available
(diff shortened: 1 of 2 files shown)


cybercop23:
I think Commit and Branch should have the banner yellow, currently green
image

OSOnly and Major should also be yellow inline with the pills - red is too scary
image

Major doesn't yet mean EoL, so let's remove that and make it a different share of yellow or orange. and the Available also yellow
image

PR #2658, opened 2026-05-25
Refactor: Bootstrap Table data-model-first multisync rewrite

jnovack:
# Refactor: Bootstrap Table data-model-first multisync rewrite

Replaces the hybrid DOM-write + `syncDOMToModel` approach in `multisync.php` with a clean Bootstrap Table data-model-first flow. All state now lives in the BT data array; `rowStyle` and `rowAttributes` apply on every render from model data.

## Architecture changes

- **BT init**: `idField: '_id'`, `uniqueId: '_id'`, `data: []` — table starts empty, data loaded via `bootstrapTable('load', systemsData)` after monkey-patch is installed
- **Monkey-patched `initBody`**: preserves warning/log child rows across re-renders; removed `syncDOMToModel` entirely
- **`parseFPPSystems`**: builds a `systemsData[]` JS array instead of HTML string rows; child rows (`_warnings`, `_logs`) still appended to DOM directly as they are outside the BT model
- **`getFPPSystemStatus`**: updates model items via `getRowByUniqueId` + direct property mutation; single `bt.initBody()` at end of batch
- **All non-FPP status handlers** (WLED, Genius, Baldrick, Falcon V3/V4, ESPixelStick): converted from direct jQuery cell writes (`$('#rowID_status').html(...)`, `$('#advancedViewUtilization_...').html(...)`) to BT model updates + `initBody()`
- **`RefreshStats`**: reads `_dataIp`, `_typeIdHex`, `_isFPP`, `_versionStr` from BT model instead of DOM; `seenRows` dedup prevents double-polling dual-IP devices on refresh
- **Deleted**: `syncDOMToModel`, `findBTItemByRowId`, `setBTColorData`, `reloadMultiSyncPage`, `closeAllLogs`, `var ipRows`, `var hostDescription`

## Deduplication

- **UUID-based**: primary dedup key; one UUID = one physical device = one row regardless of how many IPs `multiSyncSystems` lists
- **Hostname+mode fallback**: for devices with empty UUID (e.g. older FPP without UUID reporting), deduplicates by `hostname + fppModeString` — eliminates the version-string mismatch that broke the original `hostKey` approach
- **IP fallback**: last resort for truly unnamed devices
- Duplicate IPs merged into primary row's `_baseIpHtml` and `_dataIplist`; only the primary IP is polled

## Wifi icons

- **Path 1** (interfaces-based): scans `data.interfaces`; shows icon next to whichever IP in the row is on an active wifi interface (`link > 0`) — covers both "polled IP is wifi" and "polled IP is eth0 but wlan0 IP is also displayed" cases (e.g. FPPv4-11 polled via eth0, icon on wlan0)
- **Path 2** (legacy `rssi` field): fires only if path 1 produced no icon
- **WLED**: wifi icon placed in IP column from `data.wifi.rssi` / `data.wifi.signal`; RSSI row removed from utilization
- Icon placement: `String.replace(ip + '</a>', ip + '</a>' + icon)` — inserts after the correct IP's anchor

## Secondary IPs

- `_extraIpHtml`: IPs from `advancedView.IPs` not present in `multiSyncSystems` (e.g. a secondary-subnet NIC) appended after wifi icon; link-local `169.254.x.x` addresses skipped

## Uptime formatting

- `formatUptime(seconds)` helper: returns `{ short, since }` — short is `Xd Yh` or `Xh Ym` (no zero-padding, 2 fields only), since is `YYYY-MM-DD HH:mm:ss` shown on hover
- FPP `advancedView.Utilization.Uptime` always arrives as a string (`"H:M"` or `"D days H:M"`); regex parser converts to seconds before formatting
- Replaces five copies of the same days/hours/minutes floor-division block across all device type handlers

## Tooling

- `SD/fpp_debug_bundle.sh`: runs on the FPP master, collects `multiSyncSystems.json`, `systemStatus.json`, and `settings.json` into a zip for offline debugging without repeated deploys

image

### DIFF
--- a/tests/playwright/tests/multisync.spec.ts
+++ b/tests/playwright/tests/multisync.spec.ts
@@ -1,0 +1,577 @@
+import { test, expect } from './fixtures';
+import type { Page } from '@playwright/test';
+
+// ============================================================
+// Mock data — planet-themed synthetic FPP network
+//
+// Six devices chosen to exercise distinct rendering paths:
+//   JUPITER  — local Pi 4, channelInputsEnabled, master branch (up-to-date)
+//   SATURN   — SanCloud BeagleBone Enhanced, master (up-to-date), warning row
+//   MARS     — remote Pi 4, feature-rgb-output branch (behind), purple bg, syncing
+//   VENUS    — Pi 3 B, player+multisync, v9.5-stable (behind), green bg, HostDescription
+//   MERCURY  — dual-NIC Pi 4, develop branch (up-to-date), wifi icon; appears twice in
+//              MOCK_SYSTEMS (one per IP) with the same uuid so parseFPPSystems() deduplicates
+//              them into a single row showing both IPs
+//   PLUTO    — typeId 0 (Unknown); never polled; shows only "Last Seen" forever
+// ============================================================
+
+const MOCK_SYSTEMS = {
+    systems: [
+        // JUPITER — local Pi 4, channelInputsEnabled, master branch, up-to-date
+        {
+            address: '192.168.0.13', hostname: 'JUPITER', local: 1,
+            fppMode: 2, fppModeString: 'player',
+            typeId: 13, type: 'Raspberry Pi 4', model: 'Raspberry Pi 4 Model B Rev 1.1',
+            version: '10.x-master-568-gaaa1111bb', majorVersion: 10, minorVersion: 1000,
+            uuid: 'M1-10000000eaa2d726', channelRanges: '0-7',
+            channelInputsEnabled: true, channelOutputsEnabled: false,
+            multiSyncCapable: 1, multisync: false,
+            lastSeen: 1780749445, lastSeenStr: '2026-06-06 07:37:25', HostDescription: '',
+        },
+        // SATURN — SanCloud BeagleBone Enhanced (typeId 70 = BeagleBone range),
+        //          master branch, up-to-date, both channel directions enabled, warning row
+        {
+            address: '192.168.0.14', hostname: 'SATURN', local: 0,
+            fppMode: 2, fppModeString: 'player',
+            typeId: 70, type: 'SanCloud BeagleBone Enhanced', model: 'SanCloud BeagleBone Enhanced',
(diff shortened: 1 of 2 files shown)


cybercop23:
Proper case Commit and Branch?

OnlineDynamic:
looks promising - can you fix merge conflicts and I'll bring it in - there is a ticket on channel range display on multisync which could also be addressed in this PR if you fancied sorting that one ;)

patdelaney:
@jnovack can you update the tile of this to be:
Refactor: Bootstrap Table data-model-first multisync UI page rewrite.

When I saw the title in the ticket update in email I thought someone was re-writing the Multisync functions ....  ughhhhh

jnovack:
> looks promising - can you fix merge conflicts and I'll bring it in - there is a ticket on channel range display on multisync which could also be addressed in this PR if you fancied sorting that one ;)

What is that one?  Do you want to get this in and then we can solve that?

Also, I rebased and repushed.  PLEASE do testing again since I've rebased.

OnlineDynamic:
The table refresh is breaking focus when trying to select filters... turn on auto refresh and try selecting platform and mode from dropdown filters, if you don't select a mode quickly the filter focus goes back to platform

OnlineDynamic:
reorder mode is also broken - I can can order but on refresh the order is lost

OnlineDynamic:
also if you tick a checkbox to select a system to carryout an action - the auto refresh clears the tick before you can perform the action

jnovack:
@OnlineDynamic 

I feel the three things are fixed, and with tests.  I don't have an interesting environment so I can't really test, I have to mock everything.

@patdelaney Can you test the things that Stuart called out?

patdelaney:
I did a first pass on Sunday when I was at Frozen Mini... and it looked good, I'll run through it tonight after work

OnlineDynamic:
@patdelaney - you happy for us to PR this in?

patdelaney:
yes  we need more eyes on it  Thanks

PR #2663, opened 2026-05-27
Update command to check for 64-bit architectures

jessica12ryan:
This minor change fixes the troubleshooting command for architecture when installed on an OS using the FPP script.
It now checks for amd64 or arm64 instead of just arm64.
Previously, the troubleshooting command would report 32-bit when ran on 64-bit systems

Plus a couple typo fixes in FPP_Install

### DIFF
--- a/SD/FPP_Install.sh
+++ b/SD/FPP_Install.sh
@@ -592,7 +592,7 @@ install_base_packages() {
             sleep 5
         fi
         
-		echo "FPP - Removing anything left that wasn't explicity removed"
+		echo "FPP - Removing anything left that wasn't explicitly removed"
 		apt-get -y --purge autoremove
 
 		echo "FPP - Updating package list"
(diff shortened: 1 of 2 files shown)


darylc:
@jessica12ryan I realised yesterday that I probably should have done that 32/64 bit check differently and perhaps not at all, since the system-upgade.php page shows it:
image

Its getting that from /api/system/status

image

How that that look on your system (that I presume is a debian pc v's installed from an actual FPP image)?

If your system does not show it properly on the about page then perhaps we do keep this and take your PR.

jessica12ryan:
image

ghormann (approved):

darylc:
@jessica12ryan @ghormann Since JessicaRyan's FPP shows the 64 bit correctly in the new about page, I'm thinking of just removing this section that I recently added into the troubleshooting commands all together rather than having a conflicting mechanism to detect 32 vs 64 bit.  Any objections?

jessica12ryan:
@darylc No objections here... I've updated my PR to remove it

PR #2677, opened 2026-06-12
Enhancement: Add mass selectable items to playlist editor

jessica12ryan:
Let's get some feedback on this. The PR is ready to go if it's accepted.

This PR introduces mass selecting items from a playlist for the purpose of removing or duplicating.

https://github.com/user-attachments/assets/c0eccfd9-c1f4-4ebf-9a1f-b03aacaa82cb

### DIFF
--- a/www/css/fpp-dark.css
+++ b/www/css/fpp-dark.css
@@ -152,6 +152,14 @@
     background-color: #1c3a6e;
 }
 
+[data-bs-theme="dark"] .playlistSelectMode .playlistSelectedEntry {
+    background-color: transparent !important;
+}
+
+[data-bs-theme="dark"] .playlistSelectActionsDropdown .buttons {
+    background-color: #198754;
+}
+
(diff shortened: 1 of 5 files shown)


jessica12ryan:
I did make one minor change. After choosing Remove Selected or Duplicate Selected, you will now be taken out of the selection view and returned to the normal playlist editor so you can quickly return to arranging your playlist and also to make the unsaved changes button appear to defer someone from immediately closing the window afterwards and forgetting saving.

This PR also fixes the issue you would face if you have a command or effect setup in your playlist, and you want to execute it multiple times. Instead of adding that command or effect and configuring it multiple times, it can simply be duplicated after being added the first time.

Also works well on mobile

jessica12ryan:
Also added "Copy to..." to enable the ability to copy items from one playlist to another
https://github.com/user-attachments/assets/9116afe5-d9a9-4649-92ec-396c72cb5f89

PR #2685, opened 2026-06-16
Add (advanced) option to enable Pi 5 Fan Header in Settings>System

### LINKED ISSUE #2670: Allow dedicated fan control on pi 5 that have dedicated fan jst connector
the new PI5 have a dedicated fan connector that is controlled diffirently to the GPIO pins 

the fan use this command 

dtparam=cooling_fan=on
dtparam=fan_temp0=40000 # temp seting - 40,000 = 40 Degreees C - so 30,000 would be 30C
dtparam=fan_temp0_hyst=10000 # this im un sure of, possibly a time to ensure its been at that temp for 10 seconds or more ?
dtparam=fan_temp0_speed=255 #100% fan speed - 0-255 for 0-100%

is there a way to detected that FPP in installed on a PI 5 and change the options ( use gpio 14 for fan control) 
to one that allow fan control, then asks whether using default fan JST connector of GPIO 14 for fan control ? 

so every install on pi5 has the options so i dont have to always find the details and modify the config.txt firmware file myself every time i modify or upgrade

jessica12ryan:
## Add Raspberry Pi 5 Dedicated Fan Header Support

The Raspberry Pi 5 has a dedicated 4-pin JST fan connector controlled via `dtparam=cooling_fan=on` in `config.txt`, which is separate from the existing GPIO 14 fan overlay (`dtoverlay=gpio-fan`). This PR adds native Pi 5 fan control to FPP's settings UI so users don't need to manually edit `/boot/firmware/config.txt`.

### Changes

**`www/settings.json`**
- Added 4 new settings in the System tab: `Pi5Fan` (checkbox), `Pi5FanTemperature` (number), `Pi5FanSpeed` (number), `Pi5FanHysteresis` (number)
- All settings use `"variants": {"Raspberry Pi": ["Pi 5"]}` so they only appear on Pi 5 hardware
- Temperature/speed/hysteresis are child settings shown only when Pi5Fan is enabled

**`www/common/settings.php`**
- Added `SetPi5FanProperties()` function that writes/removes the following lines to `/boot/firmware/config.txt`:
  - `dtparam=cooling_fan=on`
  - `dtparam=fan_temp0=<temp_millidegrees>`
  - `dtparam=fan_temp0_speed=<speed_0_255>`
  - `dtparam=fan_temp0_hyst=<hysteresis_millidegrees>`
- Has a Pi 5 hardware guard (reads `/proc/device-tree/model`)
- When disabled, lines are commented out with `#` prefix (matching existing GPIO fan pattern)
- Added dispatch cases in `ApplySetting()` for all 4 settings

**`scripts/handle_boot_actions`**
- Added `Pi5Fan` boot-time application alongside the existing `GPIOFan`

### UI Preview

In Settings → System on Pi 5, users will see:

```
[ ] Pi 5 Fan Header
    Fan Temp Target: [50] C
    Fan Speed:       [255]
    Fan Hysteresis:  [10] C
```

UI level must be set to Advanced or higher.

### Backward Compatibility

- Existing GPIO 14 fan control (`GPIOFan` / `GPIOFanTemperature`) is completely unchanged
- Pi 5 fan uses a different mechanism (`dtparam` vs `dtoverlay`) — no conflict
- Settings default to disabled — opt-in only
- All new settings require a reboot to take effect (`"reboot": 1`)
- Code follows the exact same patterns as the existing `SetGPIOFanProperties()` function

### Additional Comments

This PR Closes #2670.

### DIFF
--- a/scripts/handle_boot_actions
+++ b/scripts/handle_boot_actions
@@ -21,6 +21,8 @@ function HandleSettings()
         ApplySetting("ForceHDMIResolutionPort2", $settings["ForceHDMIResolutionPort2"]);
     if (array_key_exists("GPIOFan", $settings))
         ApplySetting("GPIOFan", $settings["GPIOFan"]);
+    if (array_key_exists("Pi5Fan", $settings))
+        ApplySetting("Pi5Fan", $settings["Pi5Fan"]);
     if (array_key_exists("Kiosk", $settings))
         ApplySetting("Kiosk", $settings["Kiosk"]);
     if (array_key_exists("EnableBBBHDMI", $settings))
--- a/www/common/settings.php
+++ b/www/common/settings.php
@@ -283,6 +283,65 @@ function SetGPIOFanProperties()
     }
 }
 
+function SetPi5FanProperties()
+{
+    global $settings;
+
+    if (strpos($settings['SubPlatform'], "Raspberry Pi 5") === false &&
+        strpos($settings['SubPlatform'], "Raspberry Pi Compute Module 5") === false) {
+        return;
+    }
+
+    $fanOn = ReadSettingFromFile('Pi5Fan');
+    $fanTemp = ReadSettingFromFile('Pi5FanTemperature');
+    $fanSpeed = ReadSettingFromFile('Pi5FanSpeed');
+    $fanHysteresis = ReadSettingFromFile('Pi5FanHysteresis');
+
+    if ($fanTemp == '') { $fanTemp = '50'; }
+    if ($fanSpeed == '') { $fanSpeed = '255'; }
+    if ($fanHysteresis == '') { $fanHysteresis = '10'; }
+
+    $pfx = "";
+    if ($fanOn == '0') {
+        $pfx = "#";
+    }
+
+    $tempMilli = $fanTemp * 1000;
+    $hystMilli = $fanHysteresis * 1000;
+
+    $lines = [
+        "dtparam=cooling_fan=on",
+        "dtparam=fan_temp0=" . $tempMilli,
+        "dtparam=fan_temp0_speed=" . $fanSpeed,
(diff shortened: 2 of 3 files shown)


dkulp:
Is any of this actually needed?  On a Pi5, the fan is enable by default (/sys/class/thermal/cooling_device0).   It's already wired into the proper thermal zones with temperature settings. (/sys/class/thermal/thermal_zone0)

jessica12ryan:
Yes. The PR still adds value for FPP:

 - FPP's OS image may strip or not include the default Pi 5 fan config — it ships its own config.txt which does not include any dtparam=cooling_fan=on or fan_temp* lines. The [pi5] section only sets gpu_mem=256 and dtparam=uart0=on. Without this PR, a Pi 5 running FPP could have no fan control at all unless the user manually adds it.
 - Customization from the UI — even when defaults are present, users currently need to SSH in and hand-edit config.txt to change the temperature threshold, speed, or hysteresis. This PR makes those tunables accessible from the FPP settings page alongside the existing GPIO 14 fan control.
 - Consistency — FPP already has "GPIO 14 Fan Control" for older Pis. This fills the equivalent gap for Pi 5's native header, using the same UX pattern.

Without this PR, every Pi 5 FPP installation boots with zero fan configuration in its config.txt — users must manually edit it after every install or upgrade. The dtparam=cooling_fan=on defaults from stock Pi OS are irrelevant because FPP generates its own config.txt from scratch.

dkulp:
The fan control is default on even without  dtparam=cooling_fan=on.    That's not needed.  It already works without it.

If we are going to add fan/temperature control, it needs to be more generic than Pi5 specific and operate on the appropriate setting values in the appropriate   /sys/class/thermal/thermal_zone# directory.  There are other capes/hats/etc... that implement PWM fans and any settings should be more generic.   It wouldn't involve config.txt settings at all.

PR #2689, opened 2026-06-19
Fix multiple minor bugs

jessica12ryan:
Six small, non-breaking fixes in the PHP web UI. No impact on FPP daemon or core functionality.

---

### Fix 1: `www/restartRemoteFPPD.php` — `$mode` used before assignment

`$mode` was referenced in an `echo` statement before it was assigned. Changed `$mode` to `htmlspecialchars($_GET['mode'])` to eliminate the PHP warning and ensure the correct value is displayed.

```diff
 if (isset($_GET['mode'])) {
-    echo "Setting FPPD mode @ " . htmlspecialchars($ip) . " to " . $mode . "\n";
+    echo "Setting FPPD mode @ " . htmlspecialchars($ip) . " to " . htmlspecialchars($_GET['mode']) . "\n";
     $mode = $_GET['mode'];
     ...
 }
```

---

### Fix 2: `www/remotePush.php` — Undefined variable `$fname`

`$fname` was never defined anywhere in the file or its includes. It was a typo for `$filename` (defined on line 42).

```diff
-    echo "Error sending" . $fname . " " . curl_error($ch);
+    echo "Error sending" . $filename . " " . curl_error($ch);
```

---

### Fix 3: `www/healthCheckHelper.php` — Unescaped `$_GET['timestamp']` in `system()`

`$_GET['timestamp']` was concatenated directly into a `system()` call with no sanitization, allowing command injection. Wrapped with `intval()` to restrict to numeric input.

```diff
-system($settings["fppDir"] . "/scripts/healthCheck --php " . $_GET['timestamp']);
+system($settings["fppDir"] . "/scripts/healthCheck --php " . intval($_GET['timestamp']));
```

Consistent with `healthCheckSSE.php:12` which already uses the same `intval()` pattern for the identical parameter. The downstream bash script (`scripts/healthCheck`) only uses the value in integer arithmetic (`$((...))`).

---

### Fix 4: `www/js/fpp.js` line 1407 — `nodename` → `nodeName`

```diff
-					if (outputArea.nodename != 'PRE' && raw == false) {
+					if (outputArea.nodeName != 'PRE' && raw == false) {
```

`nodename` is not a valid DOM property — the correct property is `nodeName` (capital N). Since `nodename` evaluates to `undefined`, the condition `undefined != 'PRE'` is always `true`. This means newlines are always replaced with `<br>` tags regardless of the target element, even inside `<PRE>` blocks where they should be preserved.

### Fix 5: `www/js/fpp.js` line 5498 — `playerStatusTest` → `playerStatusText`

```diff
-			playerStatusTest =
+			playerStatusText =
 				"<strong>'" + jsonStatus.current_sequence + "'</strong>";
```

`playerStatusText` is declared on line 5494 and used on line 5507 to update the DOM (`$('#txtPlayerStatus').html(playerStatusText)`). However, the sequence name (wrapped in `<strong>`) was assigned to a typo'd variable `playerStatusTest` on line 5498 — which is never read anywhere. The sequence name is silently dropped from the player status display.

---

### Fix 6: `src/channeloutput/PCA9685.cpp` lines 179 and 182 — `==` used instead of `=`

```diff
         if (dt == PCA9685Output::DataType::ABSOLUTE_DOUBLE) {
             z /= 2.0f;
-            dt == PCA9685Output::DataType::ABSOLUTE;
+            dt = PCA9685Output::DataType::ABSOLUTE;
         } else if (dt == PCA9685Output::DataType::ABSOLUTE) {
             z *= 2.0f;
-            dt == PCA9685Output::DataType::ABSOLUTE;
+            dt = PCA9685Output::DataType::ABSOLUTE;
         }
```

`==` performs a comparison and discards the result — a no-op. The variable `dt` is never changed, so the downstream `if (dt == ABSOLUTE)` block (which applies the PWM frequency scaling formula at lines 184-191) never executes for `ABSOLUTE_DOUBLE` input. Values are silently output at roughly half their intended magnitude with no scaling applied.

### DIFF
--- a/src/channeloutput/PCA9685.cpp
+++ b/src/channeloutput/PCA9685.cpp
@@ -176,10 +176,10 @@ unsigned short PCA9685Output::PCA9685Port::readValue(unsigned char* channelData,
             DataType dt = m_dataType;
             if (dt == PCA9685Output::DataType::ABSOLUTE_DOUBLE) {
                 z /= 2.0f;
-                dt == PCA9685Output::DataType::ABSOLUTE;
+                dt = PCA9685Output::DataType::ABSOLUTE;
             } else if (dt == PCA9685Output::DataType::ABSOLUTE) {
                 z *= 2.0f;
-                dt == PCA9685Output::DataType::ABSOLUTE;
+                dt = PCA9685Output::DataType::ABSOLUTE;
             }
             if (dt == PCA9685Output::DataType::ABSOLUTE) {
                 z *= frequency;
--- a/www/healthCheckHelper.php
+++ b/www/healthCheckHelper.php
@@ -6,5 +6,5 @@ require_once ("common.php");
 
 DisableOutputBuffering();
 
-system($settings["fppDir"] . "/scripts/healthCheck --php " . $_GET['timestamp']);
+system($settings["fppDir"] . "/scripts/healthCheck --php " . intval($_GET['timestamp']));
 ?>
--- a/www/js/fpp.js
+++ b/www/js/fpp.js
@@ -1404,7 +1404,7 @@ function StreamURL (
 					outputArea.nodeName == 'PRE' ||
 					outputArea.nodeName == 'SPAN'
 				) {
-					if (outputArea.nodename != 'PRE' && raw == false) {
+					if (outputArea.nodeName != 'PRE' && raw == false) {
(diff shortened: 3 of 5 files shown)


darylc:
@jessica12ryan I'll accept this but we really prefer unrelated commits to be in their own PR in future please.

PR #2702, opened 2026-06-23
Bugfix: Filter out fan speed sensors from sensor updates

jessica12ryan:
This PR is a minor bug fix for a bug introduced from PR #2699.

The fan temp takes over the CPU temp on the header bar (see screenshot).
Screenshot 2026-06-23 at 9 59 52 AM

The CPU temperature should be displayed with fan info in the tool tip (fix - see screenshot).
Screenshot 2026-06-23 at 10 00 35 AM


dkulp:
This would filter OUT the Voltage sensors which have been there since FPP5 (or older).  I'd consider that a regression.

jessica12ryan:
@dkulp I missed that... Changed it to filter out just the fan speeds as they are displayed in system-stats

PR #2717, opened 2026-07-02
Feedback Fixes #2613 - about.php updates

### LINKED ISSUE #2613: fpp 10 about/upgrade page recommendation

agent462:
Per Feedback Fixes #2613 — https://github.com/FalconChristmas/fpp/issues/2613

## Summary

Reworks the About page upgrade experience and the changelog into a cleaner, more coordinated design, plus a set of UX refinements.

## Changes

- **Redesigned the About upgrade UI** with a coordinated recommendation layout so the recommended action reads clearly.
- **Un-fade the OS card and button** once a user selects an image, and show an **empty-state note** when no OS image matches the current platform.
- **Gate nightly and other prereleases separately** in the OS image list.
- **Added skeleton shimmer** placeholders for version fields while they populate asynchronously.
- **Restyled the FAQ** as a flat divider list and fixed the reveal animation to drive `max-height` from JS instead of grid rows.
- **Redesigned the changelog** as a commit list with a revert modal.
- Normalized CSS formatting in `fpp-system-design.css`.

### DIFF
--- a/www/about.php
+++ b/www/about.php
@@ -1423,7 +1423,7 @@
                         <h4 class="fpp-info-panel__title"><i class="fas fa-question-circle"></i> Frequently Asked
                             Questions</h4>
                         <div class="fpp-faq" id="upgradeFaq">
-                            <div class="fpp-faq__item">
+                            <div class="fpp-faq__item fpp-faq__item--open">
                                 <div class="fpp-faq__question">
                                     Which upgrade should I choose?
                                     <i class="fas fa-chevron-down"></i>
--- a/www/about.php
+++ b/www/about.php
@@ -1026,13 +1026,29 @@
         }
 
         function initFaqAccordion() {
-            document.querySelectorAll('.fpp-faq__item').forEach(item => {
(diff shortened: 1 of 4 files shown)


jessica12ryan:
I like the fading of the upgrade sections, but if I may make one minor suggestion, I think these grey dots should be green. Green = good... the grey makes me think at first glance that it cant reach the network to check for updates.
fppupdate

PR #2741, opened 2026-07-16
Show missing sequence/media indicators inside playlist editor

### LINKED ISSUE #2738: Indicator for missing sequences in PLAYLISTS
FPP version 9.5.3 running on a pi 3B+

In FPP -> PLAYLISTS - if you've removed a sequence that is linked inside a playlist, you get a big yellow exclamation point icon (which is very helpful) - and then when you drill into a playlist with one of those, it USED TO also mark the specific sequence with the same marking - so you knew which one was the issue. Version 9.5.3 isn't doing that. 

If you drill into a PLAYLIST that contains an exclamation mark, and you do find the offending sequence that's missing it's file(s), and correct that link, then click SAVE PLAYLIST, it does properly remove the yellow exclamation mark on the top-level playlist view. This issue is only INSIDE of the playlist.

NOTE: This could have been broken anywhere between 7.5 and 9.5. I was way behind in upgrading. :)

jessica12ryan:
## Summary

When a playlist entry references a missing sequence, media, playlist, or image file, the top-level playlist card view correctly shows a yellow warning indicator — but the drill-in editor view shows nothing. This restores the per-entry warning indicator inside the playlist editor so users can immediately identify which entry is broken without manually checking each one.

### `www/js/fpp.js`

- **`LoadPlaylistDetails()`** — After loading the playlist data, parses the existing validation messages from `playListArray` (fetched by `GET /api/playlists/validate`) to build a set of invalid entry names. Passes the set to `PopulatePlaylistDetails()`.

- **`PopulatePlaylistDetails()`** — Accepts a new optional `invalidNames` parameter (default `{}`, no breaking change) and forwards it to `GetPlaylistRowHTML()`.

- **`GetPlaylistRowHTML()`** — Accepts a new optional `invalidNames` parameter. When an entry references a name present in the invalid set:
  - Adds a `playlistRowWarning` CSS class to the `<tr>` (yellow background highlight)
  - Adds a `title` attribute with details (e.g. "Missing sequence: show.fseq")
  - Renders a warning icon (`⚠`) next to the entry type label in the header

### `www/css/fpp.css`

- **`.playlistRowWarning`** — Yellow background using Bootstrap's `--bs-warning-bg-subtle` variable (works correctly in both light and dark themes)
- **`.playlistEntryWarningIcon`** — Inline warning symbol styled with `--bs-warning` color

## How It Works

The backend already computes which entries are invalid via the existing `GET /api/playlists/validate` endpoint and loads the results into the global `playListArray` during page initialization. The fix simply plumbs that existing validation data into the per-entry rendering code — no new API endpoints, no backend changes, and no breaking changes to any existing function signatures (all new parameters default to `{}`).

## Verification

- New entries created in the editor don't show warnings (correct — they haven't been saved yet)
- After saving a playlist, the warning updates correctly (the save flow refreshes `playListArray` via `PopulateLists()` before re-rendering)
- The playback/non-edit view is unaffected (doesn't pass `invalidNames`, defaults to `{}`)
- Dark mode is supported via Bootstrap CSS custom properties

## Additional Comments

This PR closes #2738.

### DIFF
--- a/www/css/fpp.css
+++ b/www/css/fpp.css
@@ -1812,6 +1812,18 @@ body.is-loading #schedulerInfo .labelValue:before {
 	margin-left: 0.8em;
 }
 
+.playlistRowWarning {
+	background-color: var(--bs-warning-bg-subtle) !important;
+}
+.playlistEntryWarningIcon {
+	color: var(--bs-warning);
+	font-family: 'Font Awesome 5 Free';
+	font-weight: 900;
+	font-size: 1.1em;
+	margin-left: 0.4em;
+	vertical-align: middle;
+}
+
 .playlistCard {
 	margin-top: 0.5em;
 	cursor: pointer;
--- a/www/js/fpp.js
+++ b/www/js/fpp.js
@@ -2118,18 +2118,42 @@ function GetPlaylistDurationDiv (entry) {
 	);
 }
 
-function GetPlaylistRowHTML (ID, entry, editMode) {
+function GetPlaylistRowHTML (ID, entry, editMode, invalidNames = {}) {
 	var HTML = '';
 	var rowNum = ID + 1;
 
+	var warningClass = '';
+	var warningTitle = '';
+	if (entry.type == 'sequence' && entry.sequenceName && invalidNames[entry.sequenceName]) {
+		warningClass = ' playlistRowWarning';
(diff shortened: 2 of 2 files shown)


jessica12ryan:
Disregard mention of "feat: Add iconURL to plugin template"... mentioned wrong issue number

Issue #2339, opened 2025-08-20
Setting Configuration on First Boot Up


thekameleon:
**Is your feature request related to a problem? Please describe.**
It's an inconvenience.

**Describe the solution you'd like**
After I image FPP onto a MicroSD card.  Would it be possible to take a JSON backup configuration stick onto the image. And have FPP automatically apply the configuration? 

**Describe alternatives you've considered**
Manually doing it like we always have

**Additional context**
One scenario is upgrading a K32A-B to K32 Max.  My configuration doesn't change, just the board. Another would be mass configuring boards.

dkulp:
There is some ability to do similar things for the Pi and BB64 images.    If there is a script called "fpp_boot.sh" in the FAT partition, that will be executed at boot.  Thus, you can pretty much do all kinds of things from the script.   Also, if there is a "fpp" dir there, anything in there will be copied into /home/fpp/media (cp -a  so full traversal).    Thus, you can setup the FAT drive for some of this.

That said, it's not really well known as it's never been usable for the Beagles as the Beagles don't have the FAT partition.   Supposedly, that may change for Trixie, but I haven't dug into that enough yet.

jessica12ryan:
Does anyone have any objections to this? I have it ready to commit and send through a PR if we are in agreement with it.

https://github.com/user-attachments/assets/3e9b280b-5218-4f8b-9034-4ec401434c47

cybercop23:
That's pretty handy. How will this work/look on a mobile phone?

dkulp:
My only objection is have it at the very top like that taking up a ton of space by default.....   90% of the time (or more) it won't be used so having it up there like that is a big issue.  Maybe a button under the "Finish Setup" for "Restore from Backup" that would expose it or something.

jessica12ryan:
https://github.com/user-attachments/assets/dbcae739-693d-4d14-a8f6-68079b283b60

https://github.com/user-attachments/assets/3deeb532-0e4c-4180-ac48-c1dcd14845e8

darylc:
@thekameleon Are you in a position to test this with the latest nightly image?

thekameleon:
Sure. Ill do so this weekend.

thekameleon:
Yes. I should be able to this week.

Yahoo Mail: Search, Organize, Conquer 
 
  On Fri, Jun 26, 2026 at 21:05, ***@***.***> wrote:   darylc left a comment (FalconChristmas/fpp#2339)
[...]

Issue #2714, opened 2026-06-29
FPP10 - File Copy Restore


bobreese:
When doing a FPP Backups - File Copy Backups - Restore From Remote FPP Backups Directory
The restore also puts all Other Machines Backups into the /media folder.

Restore method used.
<img width="1816" height="826" alt="Image" src="https://github.com/user-attachments/assets/c32e15a0-a5bd-48b8-9ee4-253c47385b89" />

/media directory
<img width="653" height="592" alt="Image" src="https://github.com/user-attachments/assets/0646a52d-abe0-4b0f-bbfb-e8fd7aea8f35" />

/media/FPPBeagleBlack directory

<img width="596" height="1000" alt="Image" src="https://github.com/user-attachments/assets/5a49ca30-480b-49f0-b8be-b0f0954bd490" />

OnlineDynamic:
Have made a change - please retest and close if fixed

bobreese:
Thank you....

PR #2744, opened 2026-07-16
feat(initialSetup) - Add File Copy Restore option to Initial Setup restore dialog

jessica12ryan:
# Add File Copy Restore option to Initial Setup restore dialog

## Summary

Enhances the **Restore from Backup** dialog on the initial setup page to support two restore methods: **JSON Configuration Restore** (existing behavior) and **File Copy Restore** (new, reusing the same backend logic as `backup.php#tab-fileCopy`).

As part of this work, the file-copy/backup/restore UI logic shared between `backup.php` and `initialSetup.php` was extracted into a reusable JavaScript module, eliminating ~420 lines of duplicated code and fixing several behavioral inconsistencies between the two pages.

### New file: `www/js/fpp-backup-filecopy.js` (302 lines)

A shared JavaScript module extracted from the common file-copy logic previously duplicated across `backup.php` and `initialSetup.php`. Provides configurable helpers:

- `directionChanged` — shows/hides the correct field groups (USB, host, path, etc.) based on the selected direction (TOUSB, FROMUSB, TOLOCAL, FROMLOCAL, TOREMOTE, FROMREMOTE)
- `getBackupDevices` — fetches USB devices and triggers directory listing
- `getBackupDeviceDirectories` / `getRestoreDeviceDirectories` — list backup directories on USB
- `getBackupDirsViaAPI` — fetches backup directories (local or remote via `?ip=`)
- `populateBackupDirs` — populates the path `<select>` from API data
- `getBackupHostBackupDirs` — fetches remote host backup directories AND checks rsync status
- `getRemoteHostUSBStorage` — fetches available USB storage on a remote host
- `checkRemoteHasRsyncdEnabled` — queries remote rsync service status and shows a warning if disabled
- `closeCopyDialog` — unmounts remote storage and closes the modal
- `copyDone` / `copyTimeoutError` — enables the close button on completion; fallback polling for long-running copies

Configuration object (`fppFileCopy.config`) allows each page to supply its own DOM selectors, making the module page-agnostic.

### `www/initialSetup.php`

**Restore dialog now has two modes**, selectable via a dropdown at the top of the modal:

1. **JSON Configuration Restore** (default)
   - Unchanged upload-and-restore flow using the existing PHP handler (`doRestore('all', ...)`)
   - Added informational alert: *"The JSON Configuration backup file does not contain sequences, media, plugin files, or additional files..."*

2. **File Copy Restore**
   - Direction selector: **Restore from USB**, **Restore from Local FPP Backups Directory**, **Restore from Remote FPP Backups Directory**
   - Conditionally shows/hides USB device list, remote host dropdown, remote storage selector, and backup path dropdown based on the chosen direction
   - **"What to restore"** checkboxes (Configuration, Playlists, Plugins, Sequences, Images, Scripts, Effects, Music, Videos, Virtual EEPROM) — rendered as a table matching `backup.php`'s `#CopyFlagsTable` layout
   - **Delete extras** checkbox
   - Calls `copystorage.php` with the same parameters (`direction`, `storageLocation`, `path`, `flags`, `delete`, `compress`) as `backup.php`'s `PerformCopy()`
   - Progress popup using `StreamURL` + fallback log polling

- Added `<script src="js/fpp-backup-filecopy.js">` include
- Added page-specific config block mapping DOM IDs (`#fc*` selectors) to `fppFileCopy.config`
- **Removed ~97 lines** of duplicated file-copy helper functions — replaced with one-line wrappers delegating to `fppFileCopy.*`
- Kept page-specific functions: `GetFileCopyFlags`, `PerformFileCopyRestore`, JSON restore flow, `restoreTypeChanged`, `openRestoreDialog`

### `www/backup.php`

- Added `<script src="js/fpp-backup-filecopy.js">` include
- Added page-specific config block mapping DOM IDs to `fppFileCopy.config`
- **Removed ~324 lines** of duplicated file-copy helper functions — replaced with one-line wrappers delegating to `fppFileCopy.*`
- Kept page-specific functions: `GetCopyFlags`, `PerformCopy`, JSON backup flow, HTML

**No backend changes** — all logic reuses existing `copystorage.php`, `backup.php`, and API endpoints.

### jQuery selector escaping for dotted IDs (`backup.php`)

The backup page uses HTML IDs containing dots (e.g. `id="backup.USBDevice"`). jQuery interprets an unescaped dot as a CSS class selector, so `$('#backup.USBDevice')` matches `id="backup"` with class `USBDevice` — never the actual element.

**Fix:** All dotted ID selectors in both `backup.php` and the shared module default config were escaped with `\\` (e.g. `'#backup\\.USBDevice'`), so jQuery's selector engine sees the escaped dot and matches the literal ID.

### Inconsistent rsync-disabled warning

The "Rsync server is disabled on remote host" warning appeared on `backup.php` when restoring from a remote host but was missing from `initialSetup.php`'s restore dialog.

**Root cause:** The rsync check only ran during `directionChanged`, which fires before the user has selected a host (host value is empty → check exits early). `initialSetup.php` lacked the secondary trigger that `backup.php` had via `PrintSettingSelect`'s callback mechanism.

**Fix:** Added `fppFileCopy.checkRemoteHasRsyncdEnabled($(fppFileCopy.config.host).val())` to the shared `getBackupHostBackupDirs` function. This function is called whenever the host or remote storage dropdown changes on both pages, so the rsync check always fires after a host is selected.

### Remote storage dropdown never populates on host change

On `initialSetup.php`, selecting a remote host in the restore dialog left `#fcRemoteStorage` stuck on "Default FPP Storage" because the host `<select>`'s `onchange` only called `GetBackupHostBackupDirs()` — which fetches backup directories but never touches the storage dropdown. The `getRemoteHostUSBStorage()` function (which populates the storage dropdown) was only called once during `directionChanged()`, at which point the host field was still empty, causing it to no-op.

Same bug on `backup.php`: `PrintSettingSelect`'s generated callback only called `GetBackupHostBackupDirs` on host change, never refreshing remote storage.

**Fix (initialSetup.php):** Changed `#fcHost` onchange to call `GetRemoteHostUSBStorage()` before `GetBackupHostBackupDirs()`. `getRemoteHostUSBStorage()` already chains into `getBackupHostBackupDirs` on success, so the explicit call provides a fallback when no USB storage options exist.

**Fix (backup.php):** Added `GetRemoteHostUSBStorage()` call to the `GetBackupHostBackupDirs` wrapper so it fires whenever the host changes.

### Incorrect post-restore redirect on `initialSetup.php`

After a file-copy restore completed and the user clicked "Close", the page stayed on `initialSetup.php` instead of navigating to the FPP homepage.

**Fix:** Added `window.location.href = '/'` to the close button's click handler in the restore dialog.

### Duplicate spinner CSS on `initialSetup.php`

The page defined its own `.fpp-backup-action-loading::after` pseudo-element and `@keyframes fpp-spin`, while `css/fpp.css` already provides a `.fpp-backup-action-loading:before` spinner globally. Both pseudo-elements rendered simultaneously — two spinners with different colors, sizes, and animation speeds.

**Fix:** Deleted the redundant `::after` spinner block and `@keyframes fpp-spin` from `initialSetup.php`'s inline `<style>`.

## Screenshots

ss1
ss2

## Additional Comments

This is also a fix for one of 3 issues reported in [#2339](https://github.com/FalconChristmas/fpp/issues/2339#issuecomment-4993280577).

### DIFF
--- a/www/backup.php
+++ b/www/backup.php
@@ -2125,6 +2125,7 @@ if ($skipHTMLCodeOutput === false) {
         <title><?= $pageTitle ?></title>
         <!--    <script>var helpPage = "help/backup.php";</script>-->
         <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+        <script src="js/fpp-backup-filecopy.js"></script>
 
         <?php
         $backupHosts = getKnownFPPSystems();
@@ -2162,6 +2163,97 @@ if ($skipHTMLCodeOutput === false) {
             }
             else {
                 helpPage = "help/" + helpPage;
+            }
+
+            fppFileCopy.config = {
+                direction: '#backup\\.Direction',
+                usbDevice: '#backup\\.USBDevice',
+                pathSelect: '#backup\\.PathSelect',
+                host: '#backup\\.Host',
+                remoteStorage: '#backup\\.RemoteStorage',
+                deleteExtra: '#backup\\.DeleteExtra',
+                copyText: '#copyText',
+                showUSB: '.copyUSB',
+                showHost: '.copyHost',
+                showHostDevice: '.copyHostDevice',
+                showPathSelect: '.copyPathSelect',
+                showPath: '.copyPath',
+                showBackups: '.copyBackups',
+                showCompressed: '.sendCompressed',
+                popupModalId: 'copyPopup_Modal',
+                popupCloseBtnId: 'copyPopup_ModalCloseButton',
+                copyPopupBodyId: 'copyPopup'
+            };
+
+            function BackupDirectionChanged() {
+                fppFileCopy.directionChanged();
+            }
+
+            function GetBackupDevices() {
+                $('#jsonConfigbackup\\.USBDevice').parent().closest('div').addClass('fpp-backup-action-loading');
+                fppFileCopy.getBackupDevices();
+            }
+
+            fppFileCopy.config.onDevicesLoaded = function (data) {
+                $('#jsonConfigbackup\\.USBDevice').parent().closest('div').removeClass('fpp-backup-action-loading');
+                var defaultOption = "<option value='none' selected>None</option>";
+                var options = '';
+                for (var i = 0; i < data.length; i++) {
+                    var desc = data[i].name;
+                    if (data[i].vendor != '') desc += ' - ' + data[i].vendor;
+                    if (data[i].model != '') {
+                        desc += (data[i].vendor != '') ? ' ' : ' - ';
+                        desc += ' ' + data[i].model;
+                    }
+                    desc += ' - ' + data[i].size + 'GB';
+                    options += "<option value='" + data[i].name + "'>" + desc + "</option>";
+                }
+                $('#jsonConfigbackup\\.USBDevice').html(defaultOption + options);
+                GetJSONConfigBackupDevice();
+            };
+
+            function GetBackupDeviceDirectories() {
+                fppFileCopy.getBackupDeviceDirectories();
+            }
+
+            function GetRestoreDeviceDirectories() {
+                fppFileCopy.getRestoreDeviceDirectories();
+            }
+
+            function USBDeviceChanged(toFromRemote) {
+                fppFileCopy.usbDeviceChanged();
+            }
+
+            function GetBackupDirsViaAPI(host, remoteStorageSelected, excludeRoot) {
+                fppFileCopy.getBackupDirsViaAPI(host, remoteStorageSelected, excludeRoot);
+            }
+
+            function GetBackupHostBackupDirs(remoteStorageSelected) {
+                fppFileCopy.getBackupHostBackupDirs(remoteStorageSelected);
+            }
+
+            function PopulateBackupDirs(data, excludeRoot) {
+                fppFileCopy.populateBackupDirs(data, excludeRoot);
+            }
+
+            function CheckRemoteHasRsyncdEnabled(host) {
+                fppFileCopy.checkRemoteHasRsyncdEnabled(host);
+            }
+
+            function GetRemoteHostUSBStorage() {
+                fppFileCopy.getRemoteHostUSBStorage();
+            }
+
+            function CloseCopyDialog() {
+                fppFileCopy.closeCopyDialog();
+            }
+
+            function CopyDone() {
+                fppFileCopy.copyDone();
+            }
+
+            function CopyTimeoutError() {
+                fppFileCopy.copyTimeoutError();
             }
 
             function GetCopyFlags() {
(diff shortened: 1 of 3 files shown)


darylc:
@jessica12ryan As discussed I am not super familiar with the backup code, so I asked Claude to take a look at the PR.  This is the raw feedback, I've not digested or challenged it.  Do you mind taking a look to see if any of it is valid?

## Summary comment (top-level)

> Nice idea, and I like that it genuinely reuses `copystorage.php` with no backend change.
>
> My main concern is structural: the ~250 lines of JS here are a hand-copy of `backup.php`'s
> inline file-copy code, and the copy has already dropped four things the original does
> deliberately (device-list chaining, the rsyncd precheck, the remote unmount, and the
> `GetRemoteHostUSBStorage()` call). It also re-implements the `excludeRoot` fix from #2714 by
> hand. As-is, every future fix to file-copy restore has to land in two places and a reviewer has
> to remember that.
>
> Could these move to a shared `www/js/fpp-backup-filecopy.js` (precedent: `js/fpp-filemanager.js`)
> parameterised on the target select ids, with both pages consuming it? That would also make most
> of the individual comments below disappear on their own.
>
> One blocker regardless of the above: the file-copy path never completes the setup flow (see
> comment on L480).
>
> Non-blocking, noted for the record: `&flags=`/`&path=` go out unencoded and the `<option>`
> builders interpolate API data unescaped — both match `backup.php` today, so not introduced here,
> just worth not spreading. And `getKnownFPPSystems()` (L138) now does a blocking localhost API
> call on every render of the first-boot page, for a dropdown two clicks deep, on a box where fppd
> may not be up yet — consider fetching it when the modal opens instead.

## Inline comments

**L480** — `StreamURL(url, 'fileCopyText', 'FileCopyDone', 'FileCopyTimeoutError');`
> Blocker: this path never finishes setup. The JSON restore sets `initialSetup-02=1` and redirects
> to `index.php` (L215-217) — the file copy just streams a log and stops. So after restoring
> `Configuration`, the user is still on the wizard, its in-memory `settings` are stale, and
> `/media/settings` + `config/` have been replaced underneath it. Clicking "Finish Setup" then
> writes `pendingSettings` back over the config they just restored (see `saveCompletionAndPasswords`,
> L~620). The JSON path redirects away specifically to avoid this. `FileCopyDone()` should at
> minimum force a reload, and a config restore probably warrants the same reboot prompt.

**L995** — `<select id="fcHost" ... onchange="GetBackupHostBackupDirs($('#fcRemoteStorage').val())">`
> `#fcRemoteStorage` can never populate. `fileCopyDirectionChanged()` only calls
> `GetRemoteHostUSBStorage()` `if ($('#fcHost').val())` (L288), but this select defaults to
> `value=""`, so it's always empty at that point — and this onchange never calls it either. Net:
> "Remote Storage" is stuck on "Default FPP Storage" forever.
>
> `backup.php` escapes this because `backup.Host` is a persisted setting (`PrintSettingSelect`,
> `backup.php:3569`) so it has a value at load, and it calls `GetRemoteHostUSBStorage()`
> unconditionally in both remote branches (`backup.php:2959`, `2973`). Adding it to this onchange
> should do it.

**L289** — `GetRemoteHostUSBStorage();` (FROMREMOTE branch)
> `backup.php:2977` also calls `CheckRemoteHasRsyncdEnabled(host)` here — remote file copy needs
> `Service_rsync` enabled on the far end. Without it a user with rsyncd off gets an opaque failure
> buried in the progress log instead of the actionable warning.

**L1098** — `GetBackupDevices();` in `openRestoreDialog()`
> Races the directory load. This is async, but switching to File Copy runs
> `fileCopyDirectionChanged()` → `GetBackupDeviceDirectories()` immediately, which reads
> `$('#fcUSBDevice').val()`. If this GET hasn't resolved, the select still holds
> `<option>Loading...</option>` — no `value` attribute, so `.val()` returns the string
> `"Loading..."`, which is truthy and gets you `GET api/backups/list/Loading...`.
> `backup.php:2456-2459` chains into `GetBackupDeviceDirectories()` from the done handler when the
> direction is FROMUSB, precisely to avoid this — worth copying the chain too.

**L483** — `function CloseFileCopyDialog()`
> `backup.php`'s `CloseCopyDialog()` (2323-2336) POSTs
> `api/backups/devices/unmount/<storage>/remote_filecopy?ip=<host>` before closing, to guarantee the
> remote USB is released if `copy_settings_to_storage.sh` didn't. Without it the device stays
> mounted and then won't show up in the next device list, since the API only offers unmounted
> devices (cf. the warning text at `backup.php:3111`).

**L507** — `if (text.includes('unmounted from') || text.length === 0)`
> These termination heuristics aren't in the original and I think both misfire.
> `text.length === 0` can enable the Close button mid-restore, telling the user it's done while
> rsync is still running. `unmounted from` never appears at all for FROMLOCAL — there's no unmount.
> `backup.php:2342` terminates only on `"File does not exist."`, which works because
> `copystorage.php` unlinks the feed at the end (see the comment at `copystorage.php:85`).
>
> Related: `var lastLen = 0;` (L494) is dead — the "no new data in N minutes" stall warning from
> `CopyTimeoutError` didn't come across with the rest of the function.

**L90** — `.fpp-backup-action-loading::after`
> This class already exists in `css/fpp.css:2349` as `:before`, and `common/menuHead.inc:84` loads
> `fpp.css` on every page — so both pseudo-elements apply and you get two spinners. This block and
> the `@keyframes fpp-spin` below it can just be deleted.

**L1046** — `<input type='checkbox' class='fcFlag' value='EEPROM' id='fcFlagEEPROM' checked>`
> `backup.php:3627` derives this default from `file_exists('.../config/cape-eeprom.bin')` rather
> than hardcoding `checked`. Arguably fine to always restore it, just flagging it as an unintended
> divergence rather than a decision.

**L953** — `<p ...>Upload a JSON backup file to restore configuration settings.</p>`
> The old copy ended "All current settings will be overwritten." The new info alert covers what
> *isn't* in the backup, but the destructive warning is gone — worth keeping both.

bobreese:
Would it be possible to add this to FPP 9.5.3 and push out a new image? I am thinking about the people that will be staying on 9.5.3 this year.

darylc:
@bobreese No, there are unlikely to be further FPP 9 releases, if people want new features they must use 10.

PR #2757, opened 2026-07-20
feat(audio): add monitoring for audio output card presence and corres…

OnlineDynamic:
…ponding warning


patdelaney:
What does this fix or do and how would we test it?

OnlineDynamic:
it means if you have a working config and then pull out the usb sound card it will now flag as an error in UI - previously was flooding the sys log but not telling user

PR #2766, opened 2026-07-23
fix(commands): Refresh page after fppd restart when command list has changed

jessica12ryan:
# fix(commands): Refresh page after fppd restart when command list has changed

## Issue

When fppd is restarted (e.g. after a plugin is installed or updated that adds or removes commands), the frontend continues to show the stale command list cached from before the restart. Users must manually refresh the page to see the updated commands — and they may not even know the command list has changed, leading to confusion when newly added commands don't appear in dropdowns or preset editors.

## Solution

Detect when fppd transitions from "not running" back to "running" (indicating a restart), then compare the current command list against a snapshot taken on page load. If the list differs — for example, a plugin registered or removed commands — the page is hard-reloaded automatically so the user sees the updated command list immediately.

## Changes

**`www/js/fpp.js`**

- **Globals** (`_fppdWasRunning`, `_savedCommandListHash`): Track fppd running state across status callbacks and store a JSON fingerprint of the command list captured on page load.
- **`captureCommandListHash()`**: Eagerly fetches `/api/commands` once on page load and stores the JSON-stringified result as a baseline.
- **`checkCommandListChanged()`**: Called after a restart transition; re-fetches `/api/commands` and triggers `location.reload()` if the list differs from the baseline.
- **`OnFPPDRestart()`**: New `OnSystemStatusChange` callback that detects the not-running → running transition of fppd.
- **Registration**: `OnFPPDRestart` is registered and `captureCommandListHash()` is invoked in `common_PageLoad_DOM_Setup`, alongside existing status-change subscribers.

## How it works

1. On initial page load, the current command list is captured as a hash.
2. `OnFPPDRestart` monitors each status update, tracking whether fppd was previously running.
3. When fppd transitions from not-running to running (restart detected), it fetches the current command list.
4. If the new list differs from the pre-restart snapshot, the page reloads automatically.

This covers all restart scenarios — manual restart via the UI button, external restart, crash recovery — via the existing WebSocket reconnection and HTTP polling mechanisms, without adding any new polling or timers.

### DIFF
--- a/www/js/fpp.js
+++ b/www/js/fpp.js
@@ -55,7 +55,7 @@ var _systemWarningInfo = [];
 // command list and refresh the page if it has changed (e.g. plugin
 // commands added/removed).
 var _fppdWasRunning = null;
-var _savedCommandListHash = '';
+var _savedCommandListJSON = '';
 var zebraPinSubContentTop = 0;
 var VolumeChangeInProgress = false;
 var VolumeChangeAPIInProgress = false;
@@ -8274,7 +8274,7 @@ function captureCommandListHash () {
 		url: 'api/commands',
 		async: true,
 		success: function (data) {
-			_savedCommandListHash = JSON.stringify(data);
+			_savedCommandListJSON = JSON.stringify(data);
 		}
 	});
 }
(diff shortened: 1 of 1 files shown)


darylc:
@jessica12ryan Maybe we could call the variable _savedCommandListJSON rather than _savedCommandListHash since it's not a hash?

jessica12ryan:
I'll fix that up!

jessica12ryan:
@darylc Ready to go

Issue #2782, opened 2026-07-27
FPP10 - Backups - Remote USB Flash Drive


bobreese:
Attempted to backup device GPIOtest to my PiMaster with a USB Flash Drive.
On GPIOtest I can pull down Remote Storage and see sda1 Sandisk
When I click on the sda1, there is a pause and then it returns to the Default FPP Storage.

In the FPPD.log I see the following

2026-07-27 15:03:09.420 copystorage.php(2047) [Backup] backup START: TOUSB sda1 (detail: logs/fpp_backup_filecopy.log, live)
2026-07-27 15:03:11.528 copystorage.php(2047) [Backup] ==================================================================================
2026-07-27 15:03:11.528 copystorage.php(2047) [Backup] Command: sudo stdbuf --output=L /opt/fpp/www/../scripts/copy_settings_to_storage.sh sda1 Automatic_Backups TOUSB none no no JsonBackups
2026-07-27 15:03:11.528 copystorage.php(2047) [Backup] ----------------------------------------------------------------------------------
2026-07-27 15:03:11.528 copystorage.php(2047) [Backup] mount: /tmp/smnt: fsconfig() failed: /dev/sda1: Can't lookup blockdev.
2026-07-27 15:03:11.528 copystorage.php(2047) [Backup]        dmesg(1) may have more information after failed mount system call.
2026-07-27 15:03:11.528 copystorage.php(2047) [Backup] Copying JsonBackups.... Please wait!
2026-07-27 15:03:11.528 copystorage.php(2047) [Backup] sending incremental file list
2026-07-27 15:03:11.528 copystorage.php(2047) [Backup]          11.66K 100%    0.00kB/s    0:00:00 (xfr#1, to-chk=0/48)
2026-07-27 15:03:11.528 copystorage.php(2047) [Backup] sent 13.07K bytes  received 39 bytes  26.22K bytes/sec
2026-07-27 15:03:11.528 copystorage.php(2047) [Backup] total size is 739.59K  speedup is 56.41
2026-07-27 15:03:11.528 copystorage.php(2047) [Backup] umount: /tmp/smnt: not mounted.
2026-07-27 15:03:11.528 copystorage.php(2047) [Backup] rmdir: failed to remove '/tmp/smnt': Directory not empty
2026-07-27 15:03:11.528 copystorage.php(2047) [Backup] =========================================
2026-07-27 15:03:11.528 copystorage.php(2047) [Backup] BACKUP COMPLETE
[...]

Issue #2781, opened 2026-07-27
FPP10 - Backups - USB JSON


bobreese:
In FPP Backups / Backup Configuration with a USB Flash drive installed and chosen.
Clicking Download Configuration opens a local Windows file folder not saving to the USB drive.

<img width="2025" height="963" alt="Image" src="https://github.com/user-attachments/assets/f07e7c92-1576-41eb-90b0-e95098abbdad" />

<img width="2035" height="1066" alt="Image" src="https://github.com/user-attachments/assets/d6c65a7e-df85-46cf-99b4-7878cc410242" />

jessica12ryan:
This can be assigned to me

darylc:
Closing per the original Backup's author commenting that this is working as designed - https://github.com/FalconChristmas/fpp/pull/2783#issuecomment-5103214771

bobreese:
@jessica12ryan While the USB drive can now be selected and saved as a backup device
The backups are failing to write to the USB flash drive.

Remote Host: 192.168.1.140 reported..."Device (sda1) mounted at (\/mnt\/remote_filecopy)"
 
rsync: [generator] recv_generator: mkdir "/GPIOTest-20260801New" (in remote_filecopy) failed: Permission denied (13)
*** Skipping any contents from this failed directory ***

[USB Backup Failure.txt](https://github.com/user-attachments/files/30617293/USB.Backup.Failure.txt)

PR #2783, opened 2026-07-28
fix(backups): USB backup fixes, 'none' guard, short tag fix, and JSON Backups file manager tab

jessica12ryan:
# fix(backups): USB backup fixes, 'none' guard, short tag fix, and JSON Backups file manager tab

## Issues Fixed

**#2781 - FPP10 Backups - USB JSON** — The "Download Configuration" button on the Backup Configuration page now saves directly to the configured USB device instead of prompting a browser file download dialog.

**#2782 - FPP10 Backups - Remote USB Flash Drive** — Selecting a USB device in the Remote Storage dropdown no longer reverts to "Default FPP Storage" after selection.

---

### `www/backup.php`

1. **USB-aware "Download Configuration" (#2781):**
   - Before: Clicking "Download Configuration" always triggered a browser download prompt regardless of whether a USB backup device was configured.
   - After: When a USB device is configured (`jsonConfigBackupUSBLocation` is set), the backup is saved locally and then copied to the USB device via `DoJsonBackupToUSB()`, and a success message is displayed on the page. The browser download prompt is suppressed. When no USB device is configured, the original browser download behavior is preserved unchanged.

2. **Remote Storage dropdown no longer resets (#2782):**
   - Before: `GetBackupHostBackupDirs()` unconditionally called `GetRemoteHostUSBStorage()` after fetching backup directories. This function rebuilt the entire Remote Storage dropdown HTML with `<option value='none' selected>Default FPP Storage</option>` as the default selection, resetting whatever device the user had just selected.
   - After: `GetRemoteHostUSBStorage()` is only called when the function is triggered by a host change (parameter is `'backup.Host'`), not when triggered by a remote storage selection change. This preserves the user's device selection while still refreshing the USB device list when switching hosts.

3. **"Backup to Local FPP File Manager?" checkbox:**
   - Added a checkbox with help tooltip that controls whether the backup file is retained in `/home/fpp/media/config/backups/` after being copied to USB. Default is checked (local copy retained). When unchecked, the local copy is removed via `@unlink()` after the USB copy completes.

### `www/api/controllers/backups.php`

4. **'none' USB setting root cause fix:**
   - Before: `GetAvailableJSONBackups()` checked `!empty($settings['jsonConfigBackupUSBLocation'])` which is true when the value is `'none'` (the default when no USB device is selected). This caused `DriveMountHelper('none', ...)` to attempt mounting a nonexistent device, returning phantom directory entries. The prune function (`pruneOrRemoveAgedBackupFiles`) counted these phantom entries, causing it to delete real backup files from `/home/fpp/media/config/backups/`.
   - After: Added `&& strtolower($settings['jsonConfigBackupUSBLocation']) !== 'none'` to the condition, preventing the phantom mount attempt.

5. **'none' guard applied to all vulnerable `DriveMountHelper` calls:**
   - All 5 `DriveMountHelper` calls that read from `$settings['jsonConfigBackupUSBLocation']` are now guarded against the `'none'` value (lines 427, 693, 778, 791, 841, 858 — covering `GetAvailableJSONBackups`, `RestoreJsonBackup`, `DownloadJsonBackup`, and `DeleteJsonBackup`).

### `www/common.php`

6. **Fixed pre-existing `<?` short open tag:**
   - Changed `<?` to `<?php` on line 1663. When PHP's `short_open_tag` is disabled, `<?` is treated as literal text rather than a PHP opening tag, causing the closing `}` of `PrintSettingPasswordSaved()` to never be parsed. This resulted in a parse error (`Unclosed '{'`) that prevented `backup.php` (and any page including `common.php`) from loading.

### `www/filemanager.php` & `www/js/fpp-filemanager.js`

7. **Renamed "Backups" tab to "File Copy Backups", added "JSON Backups" tab:**
   - The existing "Backups" tab showed files from `/home/fpp/media/backups/` (full system backups from the file copy feature). This is now labeled "File Copy Backups" to be more descriptive.
   - A new "JSON Backups" tab shows files from `/home/fpp/media/config/backups/` (JSON configuration backups). It uses the existing `api/files/JsonBackups` API endpoint via `GetDirSetting('jsonbackups')` and supports Download and Delete operations.
   - Tab order: JSON Backups appears first, followed by File Copy Backups.

### DIFF
--- a/www/api/controllers/backups.php
+++ b/www/api/controllers/backups.php
@@ -424,7 +424,7 @@ function GetAvailableJSONBackups(){
 	$json_config_backup_filenames = process_jsonbackup_file_data_helper($json_config_backup_filenames, $dir_jsonbackups);
 
 	//See what backups are stored on the selected storage device if it's value is set
-	if (isset($settings['jsonConfigBackupUSBLocation']) && !empty($settings['jsonConfigBackupUSBLocation'])) {
+	if (isset($settings['jsonConfigBackupUSBLocation']) && !empty($settings['jsonConfigBackupUSBLocation']) && strtolower($settings['jsonConfigBackupUSBLocation']) !== 'none') {
 		$dir_jsonbackupsalternate = GetDirSetting('JsonBackupsAlternate');
 
 		//$settings['jsonConfigBackupUSBLocation'] is the selected alternative drive to stop backups to
@@ -690,7 +690,7 @@ function RestoreJsonBackup(){
 				$restore_status['Message'] = 'Backup File ' . $fullPath . ' could not be read.';
 			}
 		} else if ((strtolower($restore_from_directory) === 'jsonbackupsalternate')) {
-			if (isset($settings['jsonConfigBackupUSBLocation']) && !empty($settings['jsonConfigBackupUSBLocation'])) {
+			if (isset($settings['jsonConfigBackupUSBLocation']) && !empty($settings['jsonConfigBackupUSBLocation']) && strtolower($settings['jsonConfigBackupUSBLocation']) !== 'none') {
 				//Mount and read the json backup from the jsonConfigBackupUSBLocation location
 				$file_contents = DriveMountHelper($settings['jsonConfigBackupUSBLocation'], 'file_get_contents', array($fullPath));
 
@@ -775,7 +775,9 @@ function DownloadJsonBackup(){
 		}
 	} elseif (strtolower($dirName) == "jsonbackupsalternate") {
 		//Use our DriveMountHelper to mount the specified USB drive and check if the file exists
+		if (isset($settings['jsonConfigBackupUSBLocation']) && !empty($settings['jsonConfigBackupUSBLocation']) && strtolower($settings['jsonConfigBackupUSBLocation']) !== 'none') {
 		$fileExists = DriveMountHelper($settings['jsonConfigBackupUSBLocation'], 'file_exists', array($fullPath));
+		}
 
 		if ($fileExists) {
 			//Content type will always be json so see the header
@@ -786,7 +788,9 @@ function DownloadJsonBackup(){
 			ob_clean();
 			flush();
 
+			if (isset($settings['jsonConfigBackupUSBLocation']) && !empty($settings['jsonConfigBackupUSBLocation']) && strtolower($settings['jsonConfigBackupUSBLocation']) !== 'none') {
 			DriveMountHelper($settings['jsonConfigBackupUSBLocation'], 'readfile', array($fullPath));
+			}
 		} else {
 			$status = "File Not Found";
 			return json(array("Status" => $status, "file" => $fileName, "dir" => $dirName));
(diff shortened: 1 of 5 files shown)


jaredb7:
>     1. **USB-aware "Download Configuration" ([FPP10 - Backups - USB JSON #2781](https://github.com/FalconChristmas/fpp/issues/2781)):**
>        
>        * Before: Clicking "Download Configuration" always triggered a browser download prompt regardless of whether a USB backup device was configured.

This was how it was originally designed. The backup goes to both locations and you get a download for your own keeping. Users would expect "Download Configuration" to give them a download in any situation.

The alternate backup is just extra spot backups can go as protection against the main FPP storage getting corrupted and the user was not downloading backups from the FPP Backup page (basically all system generated backups). Users should still be able to trigger a backup and get the a copy though.

The Fixes in 4. & 5. sound like the proper fix for #2781

Good work on the fixes though, I figured #2781 was just the USB not getting mounted :)

darylc:
I know nothing about the backup subsystem so I won't be reviewing this PR, Re www/filemanager.php & www/js/fpp-filemanager.js Please don't create another tab in the file manager though.  There are already too many and it is line wrapping even on 14" screens in basic UI mode.

jessica12ryan:
I do think JSON Configuration Backups should be accessible to users in addition to File Copy Backups.
Perhaps with FPP growing, the File Manager should maybe be improved to support a larger scale of items.
I'm just thinking out loud but let me know what you think.

I can remove the changes to the File Manager. I'll have to tidy up some comments and tooltips referencing the two backup tabs.

darylc:
@jessica12ryan I don't disagree with them being available, they just need to be in the same tab as the rest of the backups somehow

jessica12ryan:
What about a single Backups tab with a dropdown to switch between File Copy Backups and JSON Backups?

darylc:
What about moving merging the 2 backup directories into 1? (I've not thought about what impact it might have)

jessica12ryan:
We could do that too. My only concerns would be that it may confuse users if they are trying to download a backup (which one to download, which is which). We don't want users to download just a configuration file thinking they've grabbed a full backup and then finding out when they restore it.

FPP automatically generates a JSON backup on every settings change as well, so the JSON backup on a production machine will typically have 50-60 (it prunes automatically after this number, I can't remember exactly how many off hand) backups showing, and then if you put both backups together, it'll get pretty cluttered.

bobreese:
Personally (if you don't mind), I prefer to keep them separate.
That way I know exactly what I am backing up and restoring.

darylc:
@jessica12ryan Cherry picked most of this PR.  

2 remaining items:
The "Backup to Local FPP File Manager?" checkbox is a good idea, but we think it's more valuable wired into the automatic per-settings-change backup (MakeJSONBackup()) as a persisted setting it sidesteps needing to touch performBackup()'s exit;/download-suppression behavior at all. Would you be up for sending a new PR for that version of it?

File manager - I realised that increasingly FPP is having subtabs in tabs, so perhaps we could do it for backups in a new PR?

I'll close this PR out for now.

PR #2784, opened 2026-07-28
Replace Raspberry Pi model logos with scalable SVGs

heffneil:
## What

Replaces the raster **Raspberry Pi model logos** in `www/images/` with scalable **SVG** equivalents, so the board logo shown for a Pi stays crisp at any DPI / display size.

- Adds SVG versions of all 14 Pi variants and removes the corresponding PNGs — **exact 1:1 name parity**, so the change is a straight swap:
  `Raspberry_Pi_{A, B, A+, B+, 2, 3, 3A+, 3B+, 4, 5, Zero, ZeroW, Zero2, Logo}`
- Updates `www/config.php`'s per-model `$settings['Logo']` mapping to point at the `.svg` files.
- `www/config.php` was the **only** reference to these images in the tree, so nothing else needed updating.
- Other board logos (QEMU, BeagleBone, Orange Pi, etc.) are untouched.

## Notes

- The UI already renders SVGs for the header logos (`falconlogo.svg`, `fpp-logo.svg`), so an SVG in the model-logo slot works with no display changes.
- These are freshly drawn vector logos, so the look differs slightly from the previous raster images. Happy to adjust styling/colors if you'd prefer them closer to the originals.

## Testing

- Confirmed name parity and that `config.php` is the sole referrer.
- SVGs are static (no scripts / external references).


dkulp:
Personally, I think the original SVG's look way better.

heffneil:
theyre png's and they aren't readable with the black text.  the Pi 5 version is great in the current version. The rest arent good.  I just made the images basedo n feedback in the zoom room to consider using a SVG.  I didn't test in place so I can attest to them working better in this example but pi4 is very bad.

heffneil:
See Patrick's PR it is much better!

Issue #2847, opened 2026-08-22
Retaining Test Configuration


heffneil:
TL;DR:

When you test in channel outputs the settings are lost in another browser tab or another device.  I am suggesting those settings should be store on the server and read each time the pages loads vs just showing "Off" which btw you can't turn them off until you select something other than Off to go back to off.

[...]

PR #2853, opened 2026-08-25
fix(outputs): restore the pixel-string Testing dropdown from the backend

heffneil:
Partial fix for #2847 — the pixel-string dropdown specifically. Deliberately **Refs**, not **Fixes**: that issue also covers retaining settings when a test *stops* (which needs an `fppd` change), and the equivalent gaps on the other test surfaces. This is the one piece that is self-contained and web-only.

---

Open **Channel Outputs** in a second browser while a pixel-string test is running and the Testing dropdown reads **Off**. It's static markup, and nothing asks `fppd` what's running.

That's worse than a wrong label: re-selecting "Off" fires no change event, so the running test can't be stopped from the page that started it.

`fppd` already reports the answer on `api/testmode` — captured from a box running a Port Number test:

```json
{"enabled":1,"mode":"Outputs","outputs":"BBB Pixel Strings","type":1,...}
```

`co-ledPanels.php` already solves exactly this for its own dropdown in `RestorePanelTestPatternState()`. This does the same for the pixel-string dropdown, on `document.ready` — the options are static markup, so it only has to run after the select exists.

**Gated on `outputs`.** Pixel, panel and PWM tests are all the same fppd test mode and all report `mode: "Outputs"` with a numeric `type`, so an ungated restore claims another surface's test. Worth noting the existing panel restore isn't gated — with a `BBB Pixel Strings` test running, calling it sets the panel dropdown to "Panel Layout". I've left that alone here rather than widening the change; happy to follow up if you'd like it fixed.

### Verified on FPP 10 hardware

BeagleBone 64, `10.x-master`, with a Port Number test running:

- fresh page load selects **Port Number**
- panel dropdown unaffected
- selecting **Off** stops the test (`enabled: 1` → `enabled: 0`)

Also checked that a panel test and a stopped test both leave this dropdown untouched.

### Scope

43 lines added, one file, no markup and no PHP changed — two JS functions and one call from the existing `document.ready`. No behaviour change when no test is running.

### DIFF
--- a/www/co-pixelStrings.php
+++ b/www/co-pixelStrings.php
@@ -161,6 +161,45 @@ function readCapes($cd, $capes)
         return $('#PixelStringPixelTiming').val();
     }
 
+    // Which "outputs" values identify a test started from THIS page. The same
+    // fppd test mode backs the panel dropdown too, and both report mode
+    // "Outputs" with a numeric "type" - so without checking which surface owns
+    // it, a panel test would set this dropdown and vice versa.
+    function PixelTestOutputTypes() {
+        return ['BBB Pixel Strings', 'DPI Pixels'];
+    }
+
+    /**
+     * Point the Testing dropdown at whatever test is actually running.
+     *
+     * fppd knows this at all times and reports it on api/testmode, but the
+     * dropdown was static markup that always loaded as "Off" while the test kept
+     * running. Worse than cosmetic: re-selecting "Off" fires no change event, so
+     * the running test could not be stopped from here in one click - the same
+     * reasoning as RestorePanelTestPatternState() in co-ledPanels.php.
+     */
+    function RestorePixelTestPatternState() {
+        $.ajax({
+            url: "api/testmode",
+            async: true,
+            dataType: 'json',
+            success: function (data) {
+                if (!data || !data.enabled || data.mode != "Outputs" ||
+                    !data.hasOwnProperty('type')) {
+                    return;
+                }
+                // Only claim a test this page's own outputs are running.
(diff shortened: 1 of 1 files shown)


dkulp:
Closing this... implemented it "differently" but likely more correct by having the testing mode returned in the normal status json that is pushed via the web sockets.   The pages can then act on it as needed.

heffneil:
Cool by me.  I was trying to minimally impact code with this PR.

heffneil:
Confirmed working on hardware — thanks, your approach is better than mine.

Tested on a BeagleBone 64 running `10.x-master-1578-g7ff219763`, pixel-string outputs:

- Started a test via the API with a Channel Outputs tab **already open**; a freshly loaded tab shows the running pattern correctly (`type: 3` → "Pixel Count by String").
- Captured a live `/fppdws` frame: `status_name: "testing"` with the full `testMode` object alongside it, so the push is doing what it should.
- Panel dropdown correctly stays "Off" during a pixel test — the gate works.

Two notes from testing, both separate from this change:

**Leaving or loading Channel Outputs can stop a running test.** `co-pwm.php` binds `StopServoTesting` to `pagehide`, so navigating between output pages killed tests I'd started elsewhere, several times. Worth a look independently of #2847.

**A long-lived tab stopped picking up changes.** One tab I'd left open across several start/stop cycles kept showing "Off" while `testMode` was arriving correctly on a fresh WebSocket connection. Fresh tabs were always right, so this may just be my tab being worn out rather than anything real — mentioning it in case it points at the status pipeline going stale after enough churn.

Issue #2855, opened 2026-08-25
PRU ring re-seats on K32-Max only when a FalconV5 smart receiver is configured; receiver telemetry then freezes until fppd restart


Starfield3:
`BBShiftString: PRU1 ring read position re-seated` fires repeatedly at 100 fps on a K32-Max
**only when a FalconV5 smart receiver is configured**. Identical pixel count on local ports is
clean. The same receiver, cable and prop on a Falcon F16V5 is clean.

**Yes, I am over the rated frame rate** — 700 px needs ~21 ms against a 10 ms budget at 100 fps.
But the over-spec is identical in the case that works, so it cannot be what makes the difference.

| longest port | fps | smart receiver | result |
|---|---|---|---|
| 700 px | 100 | **no** (local ports, 1392 px total) | **113 min clean** |
| 700 px | 100 | **yes** | **7 faults**, mean 19 min |
| 700 px | 100 | yes — same receiver + cable, **on a Falcon F16V5** | **90 min clean** |

The third row is a straight swap: the **same receiver, the same network cable**, the same prop and
the same sequence, moved to a Falcon F16V5 — **running the same bidirectional protocol, with
per-port current, fuse state and receiver volts/temperature polled live throughout** (2559 polls,
zero telemetry freezes). So bidirectional telemetry at 100 fps with this receiver is not a
fundamental limit; it works on other hardware. Zero fuse trips, zero receiver dropouts, zero telemetry
freezes, and no visible corruption over 90 minutes of continuous 100 fps playback. Against the K32's
mean time-to-failure of 19 minutes that is ~99% against chance.

**My question:** is a re-seat expected under this over-spec, and is a smart receiver expected to
make it this much more likely — or does this point at a fault in my particular cape?

Three things look wrong regardless of the cause: the message claims the output is aligned when it
is not, receiver telemetry freezes and never recovers, and nothing self-recovers. Details below.

Happy to run diagnostics, enable debug logging, or test patches.

---

<details>
<summary><b>Environment</b></summary>

| | |
|---|---|
| FPP | 10.0 (v10.0 branch) |
| SBC | PocketBeagle 2 (AM62x) |
| Cape | K32-Max, eeprom v4.0.2, `pinoutVersion` 1.x |
| `stringsPerPin` | 8 (not declared by the cape; default) |
| Cape PRU map | all 64 outputs on PRU1, none on PRU0 |
| Receiver | Falcon SRx1 V5, `hw` 252, firmware build 4 |
[...]

Issue #2846, opened 2026-08-22
FalconV5Support: log tag passed as format string discards the JSON parse error


Starfield3:
### Summary

`FalconV5Support.cpp:408` passes the log tag in the **format** slot, so the exception text and the
offending JSON are discarded. Every V5 response-parse failure logs the single word
`FalconV5Support` with no diagnostic.

### The code

https://github.com/FalconChristmas/fpp/blob/master/src/non-gpl/FalconV5Support/FalconV5Support.cpp#L407-L409

```cpp
} catch (std::exception& e) {
    LogInfo(VB_CHANNELOUT, "FalconV5Support", "Error (%s) handling response JSON: %s\n",
            e.what(), SaveJsonToString(json, "  ").c_str());
}
```

`src/log.h:132` defines:

```c
#define LogInfo(facility, format, args...) _LogWrite(__FILE__, __LINE__, LOG_INFO, facility, format, ##args)
```

So `"FalconV5Support"` becomes the format string. It contains no conversion specifiers, so printf
emits just that word and drops the remaining arguments. This is the only `Log*` call in the file.

### Observed

FPP 10.0 (`370e62ed7`), PocketBeagle 2 + Kulp K32-Max cape v4.0, remote mode, one Falcon V5
smart receiver on port 33 (`differentialType: 10`).

915 identical lines in 13 minutes:

```
2026-08-22 16:29:57.802 fppd(91333) [ChannelOut] /opt/fpp/src/non-gpl/FalconV5Support/FalconV5Support.cpp:408: FalconV5Support
2026-08-22 16:29:57.898 fppd(91333) [ChannelOut] /opt/fpp/src/non-gpl/FalconV5Support/FalconV5Support.cpp:408: FalconV5Support
...
```

### Impact

Two problems, and the second is the reason for filing:

1. The log fills with a repeated token that carries no information.
2. **V5 responses are failing to parse and there is no way to find out why.** The exception is
   caught so nothing crashes, and telemetry from the attached receiver works normally (ports
   33A-36A report current, fuse status and pixel count). The failures may simply be queries to
   differential ports with no receiver attached — but that cannot be confirmed from the outside,
   because the diagnostic is destroyed at the point it would be useful.

### Suggested fix

```cpp
LogInfo(VB_CHANNELOUT, "FalconV5Support: error (%s) handling response JSON: %s\n",
        e.what(), SaveJsonToString(json, "  ").c_str());
```

[...]

PR #2881, opened 2026-09-01
BBShiftString: drop frames the PRU can't take yet; warn when the sequence rate exceeds string capacity

Starfield3:
Follows up #2855 — opening this as requested.

## What this does

`SendData()` now declines a frame while the previous one is still pending (`pendingSeq != pumpedSeq`), and a persistent UI warning appears when the sequence rate is measurably beyond what the configured strings can output.

`pumpFrameData()` only snapshots `pendingFrame` once the firmware is ready for another frame (`command == 0 && ring drained`). A frame staged before that point is simply overwritten and never rendered. When the sequence frame rate exceeds what the string length can physically clock out, that surplus of never-rendered frames is what eventually leaves the PRU ring misaligned — the c6d86600 resync then heals the ring position, but the receiver picture stays wrong and telemetry stays frozen until fppd restarts.

Declining at the source drives the surplus to zero at whatever rate the hardware actually sustains — exactly the "fully dropping frames" behaviour you described, with no fixed divisor to mis-tune. Configurations inside their budget are untouched: nothing is pending when the next frame arrives, so nothing is declined.

`pumpedSeq` becomes `std::atomic<uint32_t>` so the output thread can read it against `pendingSeq` without a formal data race; the pump's own uses compile unchanged.

## The warning

Measured rather than computed: the declined fraction over each 6000-frame window *is* "how far past the budget you are", with no dependency on per-protocol bit timing. `WarningHolder` id 61, with hysteresis — on at ≥10% dropped, off again below 2% (e.g. when a lower-rate sequence starts):

> Sequence frame rate is higher than the configured pixel strings can output; frames are being dropped

The thresholds are judgement calls — happy to adjust.

## Evidence (K32-Max on PocketBeagle 2, FPP v10.0, FalconV5 SRx1 on 4 ports, 2002 px, longest string 700 px)

Controlled A/B on the same box, same evening, one variable: **stock failed at 3 min 31 s; the gated build ran 131 minutes clean.** Roughly six hours of clean soak since across repeated runs, `re-seated 0` throughout.

The declined fraction tracks `1 − achieved/offered` at every rate tested:

| offered | PRU-achieved | declined (measured) | 1 − achieved/offered |
| --- | --- | --- | --- |
| 40 fps | 40.0 | 0.0% | 0% |
| 66.7 fps | 45.8 | 31.1–31.2% | 31.3% |
| 100 fps | 44.0 | 55.8–56.0% | 56.0% |

Verified from the other side with a temporary counter on accepted `m_curFrame` gaps: at 40 fps every one of 6000 frames advanced exactly one step; at 66.7 the accepted-frame pattern matched the predicted 55/45 mix of 1- and 2-frame steps to three decimals. Content cadence behaves exactly as "drop the frames that don't fit" predicts.

## Notes

- The gate is inside `if (falconV5Support)` because that is the configuration where the misalignment reproduces and where the soak time is. The condition itself doesn't depend on V5 — trivially widened if you'd rather it cover all BBShiftString outputs.
- The two hunks are identical to what has been running on my show rig (v10.0) for the soaks above; anchors on master are unchanged, so this is the same code rebased.

### DIFF
--- a/src/non-gpl/BBShiftString/BBShiftString.cpp
+++ b/src/non-gpl/BBShiftString/BBShiftString.cpp
@@ -1346,6 +1346,71 @@ int BBShiftStringOutput::SendData(unsigned char* channelData) {
     if (falconV5Support) {
         falconV5Support->processListenerData();
     }
+
+    // ---- back-pressure gate --------------------------------------------------
+    // pumpFrameData() only snapshots pendingFrame once the firmware is ready for
+    // another frame (command == 0 && ring drained).  A frame staged before that
+    // point is simply overwritten and never rendered.  When the requested frame
+    // rate exceeds what the configured string length can physically clock out,
+    // that surplus of never-rendered frames is what eventually leaves the PRU
+    // ring misaligned (#2855).  Declining the frame while the previous one is
+    // still pending drives the surplus to zero at whatever rate the hardware
+    // actually sustains - frame dropping at the source, with no fixed divisor
+    // to mis-tune, and configurations already inside their budget are untouched
+    // (nothing is pending when the next frame arrives, so nothing is declined).
+    //
+    // Cannot wedge: we only decline while pendingSeq != pumpedSeq, so
+    // pumpFrameData() still enters its busy branch and its 250ms watchdog
+    // still runs.
+    //
+    // Gated to falconV5Support because that is the configuration where the
+    // misalignment was reproduced and where this fix has soak-time behind it;
+    // trivially widened later if wanted.
+    if (falconV5Support) {
+        bool taken = true;
+        if (m_pru1.maxStringLen) {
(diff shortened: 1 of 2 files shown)


Starfield3:
Thanks for the quick merge!

One follow-up: could `fc4bac547` also be cherry-picked to the **v10.0 branch**? The misalignment it guards against affects current v10.x users — v10.0 is where I reproduced and soaked all of this — so it would be good if the next point release carries the gate rather than waiting for the next major.

Same situation applies to the FalconV5Support parse-logging fix from #2846 (`139d7e6b`), which is also master-only right now.

Happy to test a v10.0 build of either on my rig.

dkulp:
Master branch is targeting 10.1 so anything there will end up as 10.1 fairly soon.

PR #2886, opened 2026-09-01
BBShiftString: warn about an over-budget frame rate as soon as it applies, not after a minute of drops

Starfield3:
Follow-up from the #2855 close discussion — the Init()-time budget warning.

## What this does

`Init()` computes the sustainable ceiling from the longest configured string — `bytes × 10 µs + 1700 µs` of measured per-frame overhead — and logs it:

```
BBShiftString: longest string 2100 bytes -> 22.7ms/frame, sustainable ceiling ~44.05 fps
```

`SendData()` re-checks whenever the published refresh rate changes (one float compare per frame otherwise), and raises/retracts warning 63 naming both numbers:

> Sequence frame rate (100 fps) is higher than the configured pixel strings can output (~44.05 fps); frames will be dropped

A 1% grace band keeps a configuration that is a hair over budget — absorbed invisibly by the back-pressure gate — from warning on every start. Retraction on a compliant rate, in `Close()`, and in the destructor, following the id-61 lifecycle.

## Why rate-change detection instead of a StartingOutput() check

My first version checked in `StartingOutput()`. Live testing killed it: the output thread keeps running across sequence transitions, so a playlist mixing frame rates never re-checks — a stale warning survived into a compliant sequence, and worse, a rate that *rises* mid-playlist would never be flagged at all. Watching for the rate change itself catches every path by construction.

## Verification (K32-Max on PocketBeagle 2, longest string 700 px = 2100 bytes)

The formula's prediction matches the measured hardware ceiling from the #2855 investigation: predicted **44.05 fps**, measured 44.0–44.2 fps (and the measured inter-frame time at saturation was 22.6–22.7 ms against the predicted 22.7 ms).

Live test, one playlist stepping through rates plus a stop/start transition:

| phase | rate | expected | observed |
| --- | --- | --- | --- |
| sequence start | 40 | no warning | none |
| mid-playlist transition | 66.7 | warning appears | "(66.67 fps … ~44.05 fps)" |
| mid-playlist transition | 100 | text updates | "(100 fps … ~44.05 fps)" |
| stop, new sequence | 40 | warning clears | none |

## Notes

- Complements the measured warning (id 61): 61 reports what actually happened over a 60-second window; 63 predicts at the first frame. Id 62 is taken by the playlist-branch warnings, hence 63.
- The 10 µs/byte assumes 800 kHz WS281x-class timing; the 1700 µs overhead is the measured number from #2855 (on this hardware the naive bit-time-only estimate picks 1-in-3 where 1-in-2 is correct).
- Not wrapped in `#ifndef PLATFORM_BBB`: the clocking physics applies to AM335x too, though the overhead constant was measured on the PocketBeagle 2 — happy to guard or adjust if you'd rather.
- The 1% grace band and the message wording are judgement calls, as ever.

### DIFF
--- a/src/non-gpl/BBShiftString/BBShiftString.cpp
+++ b/src/non-gpl/BBShiftString/BBShiftString.cpp
@@ -31,6 +31,7 @@
 
 // FPP includes
 #include "../../Sequence.h"
+#include "../../channeloutput/channeloutputthread.h"
 #include "../../Warnings.h"
 #include "../../common.h"
 #include "../../log.h"
@@ -106,6 +107,7 @@ BBShiftStringOutput::~BBShiftStringOutput() {
     // idempotent; Close() normally does this, but an output torn down without
     // one must not leave its frame rate warning stranded in the UI
     setFrameRateWarning(false);
+    clearBudgetWarning();
     BBBPru::ddrRelease("BBShiftString");
     m_pumpRunning = false;
     if (m_pumpThread.joinable()) {
@@ -615,6 +617,15 @@ int BBShiftStringOutput::Init(Json::Value config) {
         m_pru1.maxStringLen = m_pru0.maxStringLen + 1;
     }
 
+    int maxLen = std::max(m_pru0.maxStringLen, m_pru1.maxStringLen);
+    if (maxLen > 0) {
+        // 10us/byte at 800KHz plus ~1700us of measured per-frame overhead
+        // (reset latch + packet staging; measured on a K32-Max, see #2855)
+        m_frameTimeUs = maxLen * 10 + 1700;
+        LogInfo(VB_CHANNELOUT, "BBShiftString: longest string %d bytes -> %.1fms/frame, sustainable ceiling ~%.4g fps\n",
+                maxLen, m_frameTimeUs / 1000.0, 1000000.0 / m_frameTimeUs);
+    }
+
     if (!StartPRU()) {
         return 0;
     }
@@ -904,6 +915,7 @@ int BBShiftStringOutput::Close(void) {
     }
     StopPRU();
     setFrameRateWarning(false);
+    clearBudgetWarning();
     for (auto& a : m_usedPins) {
         PinCapabilities::getPinByName(a.first).releasePin();
     }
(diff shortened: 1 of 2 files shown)


dkulp:
Thanks for this — and the `StartingOutput()` story is the right call for the right reason. The output thread surviving sequence transitions means a start-time check misses both a stale warning and a rate that *rises* mid-playlist; watching the rate itself catches every path by construction. Good that you found that by testing it rather than reasoning about it.

One change I'd want before this merges, then some notes.

## The bit cell is already resolved — please use it

`m_frameTimeUs = maxLen * 10 + 1700` hardcodes 8 × 1250 ns. But `Init()` calls `resolveTiming(config)` well before this point and leaves the resolved cell in `m_lowNs` (with `m_t0Ns`/`m_t1Ns`), taken from `PixelString::protocolTiming()`. That function has eight non-ws281x entries and several are a lot slower:

| protocol | `periodNs` | implied µs/byte | effect on the predicted ceiling |
| --- | --- | --- | --- |
| ws281x family | 1120 | 10.0 | correct |
| `ucs1903`, `ws2811slow` | 2500 | ~21 | **~2.1x too high** |
| `tm1803` | 2625 | ~22 | ~2.2x too high |
| `gw6205` | 2375 | ~20 | ~2.0x too high |
| `sk6822`, `pl9823` | 1750 | ~15 | ~1.5x too high |
| `ucs1912` | 1625 | ~14 | ~1.4x too high |
| `tm1804`, `tm1809slow` | 1500 | ~13 | ~1.3x too high |
| `ge8822` | 1375 | ~12 | ~1.2x too high |

So a 400 kHz string at 40 fps — genuinely over budget, real ceiling around 21 fps at your 2100 bytes — produces no warning at all. The check goes quiet on exactly the configurations that most need it, which is worse than not having it, because an absent warning now reads as "you are inside budget".

There's a calibration wrinkle worth knowing about. `m_lowNs` is the *programmed* wait, not the achieved cell: `timingCycles()` subtracts `TIMING_OVERHEAD_CYCLES`, and the bit loop has instruction time outside the waits. Your own measurement pins the gap — 2100 bytes at 22.6-22.7 ms, less 1700 µs, is 1.250 µs/bit exactly, against a programmed 1120 ns. That implies ~130 ns/bit of un-waited firmware overhead, and `bitNs = m_lowNs + 130` reproduces your ws281x number exactly while generalising to the rest of the table.

I'd rather you confirm that adder on one port configured as `ucs1903` than take it on my arithmetic — it is inferred from a single ws281x data point. But the *ratio* is dominated by `periodNs` either way, so even an uncalibrated `m_lowNs`-derived estimate is a large improvement on the hardcoded 10.

## The 1700 µs is less constant than it looks

It's one cape, 8 deep, 64 outputs, with a receiver attached, and it bundles at least two different things.

The smart receiver's config packet is part of the frame rather than overhead outside it: `pumpFrameData()` streams `total = frameBytes + packetBytes` under a single command, and the firmware clocks those bits (`OUTPUT_FULL_BIT_FV5`, `WAITNS_LOOP FALCONV5_PERIOD`). `packetBytes` is `57 * stringsPerPru() * pi->len` — it scales with the cape's depth and pin count, and it is **zero when no receiver is configured**. The remainder is presumably the inter-frame reset/latch gap.

The direction matters: without a receiver the real overhead is smaller, the predicted ceiling is too low, and it **over**-warns. Only a few percent at long string lengths, but that is more than the 1% grace band absorbs, and it is the shape of thing that generates "why is FPP warning me, it plays fine" reports. Either widen the band to ~5% so it covers the constant's own uncertainty, or split the packet term out and compute it from `stringsPerPru()` and the packet length instead of folding it into a fixed number. I'd take the wider band — it is honest about the precision actually available.

## Notes, none blocking

- **Id 63 is free.** Confirmed against the tree; 61 and 62 are the only ones in that range. Your reasoning was right.
- **Leaving AM335x unguarded is a judgement call rather than a defect, and I'd keep it.** `libfpp-co-BBShiftString.mk` builds on all BB platforms, so this does reach AM335x, where there is no back-pressure gate and over-budget has always degraded quietly into tearing and dropped frames. The warning is true there and arguably overdue. Just be aware it will newly light up on installs that have run that way for years without anyone noticing.
- **The check sits ahead of `if (!hasStrings()) return 0;`.** Harmless, since `m_frameTimeUs` is 0 and it returns early, but it is cheaper and reads better after the guard.
- **`GetChannelOutputRefreshRate()` reads a plain non-atomic global** written by the sequence thread. Not introduced here — `channeloutputthread.cpp` already reads `RefreshRate` unsynchronised from the output thread — so it is a pre-existing pattern and not a reason to hold this up.
- **`m_lastBudgetRate` is not reset in `Init()`/`StartPRU()`.** Moot today: `ChannelOutputSetup` closes and deletes outputs on a config reload, so every `Init()` runs on a fresh object with the `-1.0f` initialiser. Only worth adding if `Init()` should ever be re-runnable on a live object.

The 61/63 split is a good pairing and you described it well — 61 reports what happened over a window, 63 predicts before the first frame reaches the hardware. Fix the bit cell and I'm happy with this.

Compiled clean here under both `-DPLATFORM_BB64` and `-DPLATFORM_BBB`.

Starfield3:
All three changes are in (`f56b6c2`): the estimate now derives from `resolveTiming()`'s `m_lowNs`, the band is 5%, and the check sits behind the `hasStrings()` guard.

And you got your ucs1903 confirmation — measured tonight on the rig. Configured the whole cape as ucs1903 (one port isn't enough; the cell the most ports want wins), fppd logged `bit timing 500/2000/2500ns`, then played a 40 fps sequence into it:

| | programmed cell | achieved cell (measured) | implied adder |
| --- | --- | --- | --- |
| ws281x (the #2855 dataset) | 1120 ns | 1245–1252 ns | +125 to +132 |
| ucs1903 (tonight) | 2500 ns | 2619 ns | **+119** |

Saturated at **21.883 fps / 45.697 ms** per frame against your predicted ~45.88 ms — 0.4% off, and the adder is protocol-independent within measurement error. I kept `+130`: it reproduces the ws281x number exactly and errs toward warning slightly early rather than slightly late.

Compiles clean here on the PocketBeagle 2 against master.

Issue #2892, opened 2026-09-03
File Copy Restore from USB backup not allowing selection of the subdirectories on the drive, only "/" (root directory)


andyaz:
FPP 10.0 64bit RPi 3 370e62ed7
OS - v2026-08 (64bit) Kernel 6.18.45-v8+

Installed from full SD Flash with RPi Imager

Was on latest 9.5. Updated before 10.0 upgrade.
Backed up to local brand new USB drive. Both File Copy and JSON. Filecopy went to /FPP-MASTER/ on USB
Flashed SD to 10.0
initial settings and reboot
restored the downloaded JSON config from my desktop.
Reboot - came up on correct WiFi network as FPP-MASTER with correct IP address and other settings.
Tried to file copy restore from the USB. The USB restore will only allow me to see "/", it does not present me with the subdirectory where the file copy went.

I verified the file copy backup did write to the USB using a windows desktop and all files are present in the sub-directories under /FPP-MASTER/

Put USB back in RPi and repowered it up. Same results. When trying to filecopy restore it does not see any of the backed up files in "/" so nothing is copied over.  I have a complete backup of my FPP-MASTER sequences, media, etc but cannot restore it from UI!

Root problem is File Copy from USB Drive does not present the option to pick a subdirectory that the backup lives in.

[FPP-MASTER_Logs_20260903_0528.zip](https://github.com/user-attachments/files/31769151/FPP-MASTER_Logs_20260903_0528.zip)

jessica12ryan:
I can't seem to replicate this. You should be able to select a subdirectory from the backup path dropdown.

<img width="1412" height="837" alt="Image" src="https://github.com/user-attachments/assets/48b11c26-169f-44e3-aa4d-3a17d0e6c919" />

<img width="1412" height="837" alt="Image" src="https://github.com/user-attachments/assets/93cfa458-4841-481f-a497-cb3b21a873fb" />

andyaz:
All I get is /  in  the pull down. It does not populate like the others.


Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

andyaz:
On closer examination I will note that this RPi has 2 usb drives installed. An SSD I use for remote backups (sdb1), and another thumb drive for the local backup for this instance (sda1). On other FPP instances there is only the one drive and they populate with the directories. 

[...]

PR #2912, opened 2026-09-06
fix(backups): correct FAT handling order, symlink preservation, and mount validation for File Copy backups

jessica12ryan:
# fix(backups): File Copy Backup failure on FAT/exFAT for Configuration and Plugins

## Summary
Follow-up to #2907 / #2903. After fixing the restore subdirectory listing race, File Copy `TOUSB` backups began correctly reporting failures (`rc=23`) for `Configuration` and `Plugins` that were previously hidden as false success. This PR fixes the underlying `rsync` failures so those backups complete on `FAT`/`exFAT`/`DOS` volumes.

## Issue
#2892 (follow-up comment https://github.com/FalconChristmas/fpp/issues/2892#issuecomment-5555487379)

After `372cb3a` (`OVERALL_RC` + `pipefail`), users with multiple USB drives reported restore UI now correctly lists subdirectories, but `TOUSB` backups to both thumbdrive (`sda1`/`FAT32`) and SSD (`sdb1`) failed when `Configuration` or `Plugins` were selected (`BACKUP FAILED rc=23`), while other areas (`Playlists`, `Sequences`, `Music`, `Videos`, etc.) succeeded.

Logs showed `sent`/`total size` with no visible error (filtered by `IgnoreWarnings()`), yet `rc=23` (partial transfer) was correctly surfaced.

## Root Cause
`scripts/copy_settings_to_storage.sh:73-77` set `EXTRA_ARGS="--no-perms"` for `FAT`/`DOS` **before** adding `-av` (`scripts/copy_settings_to_storage.sh:173` did `EXTRA_ARGS="$EXTRA_ARGS -av ..."`). Since `-a` implies `-p` (`--perms`), the early `--no-perms` was overridden and `rsync -a` still attempted to preserve permissions/ownership on `FAT`, which has no on-disk ownership (mount handles `uid`/`gid` via `uid=$FPP_UID,gid=$FPP_GID` per #2782). Resulting `failed to set permissions` / `chown` / `attrs were not` errors were filtered from output but still caused `rc=23`, now correctly reported.

`Plugins` additionally contains symlinked `.so` files (e.g. `libpiper_phonemize.so`, `libespeak-ng.so` in `fpp-plugin-BackgroundMusic` and `libfpp-*.so` plugins). `rsync -a` preserves symlinks (`-l`); `FAT`/`exFAT` cannot store symlinks, causing `symlink has no referent` / `Operation not permitted` and `rc=23`.

## Fix
`scripts/copy_settings_to_storage.sh`:

* `EXTRA_ARGS` construction reordered: `EXTRA_ARGS="-av --progress --info=name0 --human-readable --modify-window=1 $EXTRA_ARGS"` so `FAT` opts override `-a` (`scripts/copy_settings_to_storage.sh:173`).
* For `FAT`/`DOS`, expand to `EXTRA_ARGS="--no-perms --no-owner --no-group --copy-links"` (`scripts/copy_settings_to_storage.sh:74,77`):
  * `--no-owner`/`--no-group` prevents `chown` errors on `FAT` (which relies on mount-time `uid`/`gid`).
  * `--copy-links` dereferences symlinks to regular files so `FAT` can store plugin libs; restore to `ext4`/`btrfs` still works (files are usable, symlink identity is not required for `FAT` backups).
* Non-breaking cleanup for `rmdir: failed to remove '/tmp/smnt': Directory not empty` (`scripts/copy_settings_to_storage.sh:275-278`, seen in `issuecomment-5555542335` after `Copying Configuration...` and `cape-eeprom.bin`): `umount -- /tmp/smnt 2>/dev/null || umount -l -- /tmp/smnt 2>/dev/null || true` and `rmdir -- /tmp/smnt 2>/dev/null || true` suppresses noisy busy-mount race (multiple USB/file-manager holds) without affecting `OVERALL_RC`. Next `mkdir -p -- /tmp/smnt` (`scripts/copy_settings_to_storage.sh:68`) recreates it; backup success unchanged.
* Mount validation hardening (`scripts/copy_settings_to_storage.sh:82-88` after `fi`): `if ! mountpoint -q /tmp/smnt 2>/dev/null; then echo "ERROR: Failed to mount..." >&2; rmdir 2>/dev/null || true; exit 1; fi` aborts before `rsync` fills local `/tmp` (tmpfs) when `mount -o nofail` fails. Prevents `rc=23`/`No space left on device` from writing to SD card and stranded files in `/tmp/smnt` (other AI diagnosis). Non-breaking: previously continued and filled `tmpfs`/SD, now fails fast with clear error.

No changes to `FROM` logic, `TOLOCAL`/`FROMLOCAL`, or `TOREMOTE`/`FROMREMOTE` paths; non-`FAT` (`ext4`/`btrfs`) behavior unchanged for `rsync` semantics.

## Testing
* `TOLOCAL` `Configuration + Plugins` on `PF-FPP-SERVER` (`ext4` root): `BACKUP COMPLETE rc=0` (`134M` sent, `1.00` speedup).
* `TOUSB` to `FAT32` loop device (`500M` `mkfs.vfat` via `losetup`/`/dev/sdz1`):
  * `Configuration` alone: `BACKUP COMPLETE rc=0` (`74K`, no `rmdir` noise).
  * `Plugins` alone: `BACKUP COMPLETE rc=0` (`217M` with `--copy-links`, previously `rc=23`, no `rmdir`).
  * `Configuration + Plugins` together: `BACKUP COMPLETE rc=0` (`217M`, `43%` used, `NO_RMDIR_FOUND` verified via `grep -i rmdir`).
  * Full set without large `Sequences` (`Sequences` `6.7G` exceeds `500M` test volume) succeeds; `Sequences` to `500M` correctly fails with `No space left on device` (`rc=11`), not `23`.
  * Busy-mount test (hold open fd on `/tmp/smnt`): `BACKUP COMPLETE rc=0` with suppressed `rmdir`, no spurious `FAILED`.
  * Mount-fail test (`sdz9` non-existent): `ERROR: Failed to mount /dev/sdz9 to /tmp/smnt` + `exit 1` and `rmdir` cleanup, no `rsync` to local `tmpfs` (previously would fill SD).
* `php -l` clean for `scripts/copy_settings_to_storage.sh`, `www/api/controllers/backups.php`, `www/common.php`.
* Existing `www/api/controllers/backups.php` per-device mountpoint fix (#2903) and `www/common.php` `GetAvailableBackupsDevices` fixes remain included; `fppd` and `http://localhost/api/system/status` healthy.

## Risk
Low. Change is `FAT`/`DOS` only; ordering fix is strictly more correct (`-av` then `--no-*`), and `--copy-links` only affects `FAT`/`DOS` where symlinks were previously failing. No API or on-disk format changes.

## Verification
* No breaking changes; `capes/`, `config/`, `Plugin.h` interfaces untouched.
* Manual `rsync -av --no-perms --no-owner --no-group --copy-links` to `FAT` verified `RSYNC_EXIT=0` for `Configuration`; without fix `rc=23` reproduced.

### DIFF
--- a/scripts/copy_settings_to_storage.sh
+++ b/scripts/copy_settings_to_storage.sh
@@ -71,10 +71,10 @@ if [ "$DIRECTION" == "TOUSB" -o "$DIRECTION" == "FROMUSB" ]; then
     elif [[ "$FSTYPE" =~ "ext4" ]]; then
         mount -t ext4 -o noatime,nodiratime,nofail -- "/dev/$DEVICE" /tmp/smnt
     elif [[ "$FSTYPE" =~ "FAT" ]]; then
-        EXTRA_ARGS="--no-perms"
+        EXTRA_ARGS="--no-perms --no-owner --no-group --copy-links"
         mount -t auto -o noatime,nodiratime,exec,nofail,uid="$FPP_UID",gid="$FPP_GID" -- "/dev/$DEVICE" /tmp/smnt
     elif [[ "$FSTYPE" =~ "DOS" ]]; then
-        EXTRA_ARGS="--no-perms"
+        EXTRA_ARGS="--no-perms --no-owner --no-group --copy-links"
         mount -t auto -o noatime,nodiratime,exec,nofail,uid="$FPP_UID",gid="$FPP_GID" -- "/dev/$DEVICE" /tmp/smnt
     else
         mount -t ext4 -o noatime,nodiratime,nofail -- "/dev/$DEVICE" /tmp/smnt
@@ -170,7 +170,7 @@ elif [ "$DIRECTION" == "FROMREMOTE" ]; then
 
 fi
 
-EXTRA_ARGS="$EXTRA_ARGS -av --progress --info=name0 --human-readable --modify-window=1"
+EXTRA_ARGS="-av --progress --info=name0 --human-readable --modify-window=1 $EXTRA_ARGS"
 
 if [ "$COMPRESS" == "yes" ]; then
         REMOTE_COMPRESS=" -Dz "
@@ -273,8 +273,11 @@ for action in "$@"; do
 done
 
 if [ "$DIRECTION" == "TOUSB" -o "$DIRECTION" == "FROMUSB" ]; then
-    umount -- /tmp/smnt
-    rmdir -- /tmp/smnt
+    # Non-breaking cleanup: umount may race as busy when multiple USB jobs or
+    # file manager still holds /tmp/smnt; suppress noisy rmdir failure and try
+    # lazy unmount as fallback. Does not affect OVERALL_RC (backup success).
+    umount -- /tmp/smnt 2>/dev/null || umount -l -- /tmp/smnt 2>/dev/null || true
+    rmdir -- /tmp/smnt 2>/dev/null || true
 fi
 
 if [ "$DIRECTION" == "TOREMOTE" -o "$DIRECTION" == "FROMREMOTE" ]; then
--- a/scripts/copy_settings_to_storage.sh
+++ b/scripts/copy_settings_to_storage.sh
@@ -79,6 +79,14 @@ if [ "$DIRECTION" == "TOUSB" -o "$DIRECTION" == "FROMUSB" ]; then
     else
         mount -t ext4 -o noatime,nodiratime,nofail -- "/dev/$DEVICE" /tmp/smnt
     fi
+    # Non-breaking mount validation: if mount failed (nofail still returns, but
+    # mountpoint not active), abort before rsync fills local /tmp (tmpfs) and
+    # causes rc=23 "No space left" / stranded files.
+    if ! mountpoint -q /tmp/smnt 2>/dev/null; then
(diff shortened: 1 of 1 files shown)


andyaz:
Question... Will the restore of whole files/directories instead of symlinks break things later? Like plugins?

jessica12ryan:
The USB stick uses a simple file system that can't store shortcuts (symlinks). Some plugins use shortcuts for their library files, so the backup to USB was failing.

This fix makes the backup copy the actual file instead of the shortcut, so the backup now completes. When you restore from USB, you'll get regular files instead of shortcuts on your system. Plugins just need the file to exist, so they will still work correctly — it just uses a little more space.

In short: backup from a system with shortcuts → USB now works, and restore from that USB → system gives you working plugins, just without the shortcut link. Much better than no backup at all.

andyaz:
That makes sense. Even if the backup used a symlink for the plugin, the config file is restored where it belongs and when you reinstall the plugin it would resolve the symlink issue, if I understand it correctly.  I guess without going to the pain of using something like a tarball to make the file metadata agnostic to the file system it is getting backed up/stored on for the backup things will work. As you can tell I am still trying to wrap my brain around things...

Thanks for the explanation, your time, and the fix!

Issue #2854, opened 2026-08-25
FalconV4 smart receivers produce no output on capes with V5 listener support: enable pin gated on hasV5SR instead of hasFalconSR


Starfield3:
## Summary

On a cape that declares `falconV5ListenerConfig` (e.g. K32-Max), a smart receiver configured as **FalconV4 produces no output at all** — no pixels, no error, no warning. The same receiver on the same port works immediately when switched back to FalconV5.

`BBShiftString.cpp:624` configures the enable pin only when a **V5** receiver is present:

```c
if (supportsV5Listeners && hasV5SR) {
    // if the cape supports v5 listeners, the enable pin needs to be
    // configured or data won't be sent on port1 of each receiver
    PinCapabilities::getPinByName(PRU1_ENABLE_PIN).configPin("pru1out", true, "BBShiftString-Enable");
}
```

`hasV5SR` is set only by `ReceiverType::FalconV5` (line 458). Under V4 it is false, the pin is never configured, and by the comment's own description no data reaches port 1 of the receiver — which is exactly the observed symptom.

## The variable for this already exists

`hasFalconSR` covers both types (line 459) and is used correctly on both sides of the line in question:

| line | purpose | flag |
|---|---|---|
| 591 | allocate the Falcon packet memory | `hasFalconSR` ✅ |
| **624** | **configure the enable pin** | **`hasV5SR`** ❌ |
| 629 | `setupFalconV5Support()` | `hasFalconSR` ✅ |

The enable pin is a property of the **cape** — it gates data to the receiver regardless of which protocol version is spoken — so gating it on the protocol version looks unintended.

## Suggested fix

```diff
-    if (supportsV5Listeners && hasV5SR) {
+    if (supportsV5Listeners && hasFalconSR) {
```

## Steps to reproduce

1. K32-Max (PocketBeagle 2 / AM62x), FPP 10.0. Falcon SRx1 V5 receiver on port 33, dial at `A` (autodetect V1/V2/V4).
2. `differentialType: 10` (FalconV5 ×1) → **receiver works, pixels light.**
3. Change to `differentialType: 16` (FalconV4 ×1), restart `fppd` → **nothing lights.** No error in the log.
4. Change back to `10`, restart → works again.

Same receiver, same cable, same dial position throughout.

## Version provenance

[...]

PR #2915, opened 2026-09-06
BBShiftString: gate the receiver enable pin on any Falcon receiver, not just V5

Starfield3:
**TL;DR** — On a cape with V5 listener support the PRU1 enable pin is gated on `hasV5SR`. With a Falcon V4 receiver that flag is false, the pad is never configured, and port 1 of every receiver stays dark. Gate it on `hasFalconSR` instead — already computed two lines away, and already used for the V5 setup right below it.

**Fix** — one member, three lines. `Close()` mirrors the same condition.

**Tested** — two vendors, same cable and string throughout, K32-Max:

| receiver | `differentialType` | before | after |
| --- | --- | --- | --- |
| Falcon SRx1 V5 | Falcon V4 (16) | silent wire, no detection | clean |
| Genius 4-Output LR | Falcon V4 (16) | silent wire, no detection | full 4-min sequence, steady 40.0 fps, 0 resyncs, 0 log errors |

**Unchanged** — `willListen` and `addListeners()` keep using `m_hasBidirSR`; those are genuinely V5-only.

**Known gap** — v1/v2 receivers set neither flag and stay dark. Same root cause, out of scope here.

Refs #2854

### DIFF
--- a/src/non-gpl/BBShiftString/BBShiftString.cpp
+++ b/src/non-gpl/BBShiftString/BBShiftString.cpp
@@ -520,6 +520,7 @@ int BBShiftStringOutput::Init(Json::Value config) {
         hasV5SR = false;
     }
     m_hasBidirSR = hasV5SR;
+    m_hasFalconSR = hasFalconSR;
 
     int curRecPort = -1;
     for (int x = 0; x < m_strings.size(); x++) {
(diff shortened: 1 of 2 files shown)


Starfield3:
Superseded by 826802ab8 — thanks for picking it up.

Your version is wider than mine, and correctly so. Gating on `supportsV5Listeners` rather than on `hasFalconSR` also covers dumb strings sitting behind the enable line; my patch would have left those dark. And the `maxStringLen` bump is a second bug I walked straight past — a config whose receiver ports all land on PRU0 would have muxed the pad to a PRU that never starts.

Closing this.

I'll build 826802ab8 on the K32-Max from #2854 and report there. That box is the broader case: no non-V5 receiver of any kind got signal, v1/v2 included, not only send-only V4. It also has a second vendor's receiver on the differential group, so it exercises a different path than your 16-port cape did.

PR #2929, opened 2026-09-09
Add a navbar icon for plugin updates

focusedonsound:
## Summary
- Adds a navbar icon (puzzle-piece) next to the existing FPP-update icon, shown only when at least one installed plugin has an available update. Mirrors `#navbarUpdateAvail`'s existing show/hide behavior and styling.
- Clicking it opens the Plugin Manager directly on the **Updates** tab (`plugins.php?tab=updates`). `plugins.php` previously had no URL deep-link into that tab (only a `sessionStorage`-remembered "last tab" from a manual click), so this adds explicit `?tab=` support to `RestoreTopTab()` that takes priority over — and then persists into — that same session memory.
- New `GET /api/plugin/updatesAvailable` endpoint aggregates the existing `PluginHasUpdates()` check across every installed plugin. That check itself is cheap (local `git log` against already-fetched remote-tracking refs, no network), but is only as fresh as the last `git fetch` per plugin. This endpoint keeps those refs from going stale on their own via a small TTL cache (same convention/location pattern as the existing GitHub-stats cache), refreshing at most **one** stale plugin's remote per call so a page load never pays for N serial `git fetch`es.

## Test plan
- [x] Verified on a live FPP 10 (master) install: new endpoint returns well-formed `{"updatesAvailable": bool}`, existing `/api/plugin` and `/api/plugin/:RepoName` routes unaffected by the new fixed-path route.
- [x] Confirmed the per-call refresh throttle: repeated polls filled in one plugin's cache entry at a time rather than all at once.
- [x] Confirmed aggregation correctness: flipping a cached entry to `hasUpdate: true` correctly flips the aggregate `updatesAvailable` to `true`.
- [x] Confirmed `plugins.php?tab=updates` reliably opens the Updates tab regardless of prior session tab state.
- [x] Confirmed in-browser: icon shows/hides correctly via `FPP_PLUGIN_UPDATE_STATE`, correct tooltip, and click-through to the Updates tab.


focusedonsound:
Sharing a screen grab of the added plugin update icon in the header
plugin_update_notification

darylc:
@focusedonsound Claude suggests a bunch of things need to be considered before this could be merged:

 Feedback for the PR author:

  Blocking
  - Cache never invalidated: hasUpdate: true sticks for up to 6h after the user upgrades, so the icon stays lit. UpgradePlugin(),
    UninstallPlugin() and InstallPluginFromInfo() need to update/forget the entry; a plugin uninstalled and reinstalled under the same
    name inherits the old flag.
  - CheckForPluginUpdates() (the Updates tab's own check) does a real fetch but doesn't record the answer, so the icon and the tab can
    disagree.
  - git fetch runs with no timeout, no connectivity gate, and stderr unredirected — an offline show box holds a php-fpm worker on it.
    Compare timeout 60 git fetch at plugin.php:2016 and the 1s ping gate in get_remote_git_version().
  - No lock, and the cache is written only after the fetch, so N concurrent page loads (tabs, kiosks) at the TTL boundary each spawn a
    fetch of the same plugin. See pluginHeaders.php for the non-blocking flock + serve-stale pattern and the worker-starvation comment
    explaining why.
  - Cache write is a bare file_put_contents; a concurrent reader can see a torn file, which json_decode turns into an empty cache and a
    full re-seed. Use write-then-rename as WritePluginHeaderIndicatorCache() does.

  Should fix
  - openapi.json is generated (www/api/tools/build_docs.sh); don't hand-edit it. Trim the docblock and regenerate — the current
    description is the whole internal comment verbatim, including "see CheckPluginsForUpdates()'s comment above".
  - Duplicates the installed-plugin directory scan; InstalledPluginNames() already does it.
  - Docblock says PluginHasUpdates() is "no network" — it also runs the plugin's scripts/fpp_update_check.sh, which may not be.
  - A failed fetch is stamped ts=now, so an outage costs 6h before retry; a shorter retry interval for failures is better.
  - Help page: CLAUDE.md requires a help update for new user-facing controls — add a sentence to www/help/plugins.php (what the icon
    means, how often FPP contacts plugin hosts).

  Minor
  - Poll fires on every page even when no plugins are installed; gate the setTimeout in menu.inc on a */pluginInfo.json existing.
  - All entries get ts=now on first check and expire together; harmless with the one-per-call cap, just means the last plugin can be
    N×6h late.
  - PR was cut against an old base (hunk offsets are ~900 lines off) — worth a rebase.

focusedonsound:
Thanks for running this through review - went through every item against the actual code before changing anything. Pushed fixes for what held up:

**Fixed:**
- Cache invalidation: `UpgradePlugin()`, `UninstallPlugin()`, and `InstallPluginFromInfo()` now all update/drop the plugin's `pluginUpdates` cache entry on success, so the icon doesn't stay lit for up to 6h post-upgrade and a reinstall under the same name starts clean.
- `CheckForPluginUpdates()` (the Updates tab's own on-demand check) now writes its result back into the same shared cache, so the tab and the navbar icon can't disagree.
- Added the same non-blocking-lock + serve-stale pattern `pluginHeaders.php`'s `GetPluginHeaderIndicators()` already uses, plus write-then-rename cache writes - concurrent page loads at a TTL boundary no longer each spawn a `git fetch` of the same plugin, and a reader can't see a torn cache file.
- The `git fetch` is now gated behind the same 1s connectivity probe `get_remote_git_version()` uses in `common.php` (not `plugin.php:2016` - that line/pattern doesn't exist; the real reference is in `common.php`, and it checks out as an accurate, directly-applicable pattern), and wrapped in `timeout 20`. A failed fetch now retries after 5 minutes instead of waiting out the full 6h TTL.
- `PluginHasUpdates()`'s "no network" characterization was wrong (it runs the plugin's own `scripts/fpp_update_check.sh`, which can do anything) - corrected, and moved the implementation-rationale comment out of the `/** */` docblock so it stops leaking verbatim (including an internal cross-reference) into the public `openapi.json`.
- `menu.inc` no longer schedules the poll at all on a box with zero plugins installed.
- Rebased onto current master and regenerated `openapi.json` against the merged state (70 commits behind, hence the stale hunk offsets).

**Pushed back on:**
- `InstalledPluginNames()` doesn't exist anywhere in this codebase - nothing to dedupe against. There's a similar-but-different scan in `pluginHeaders.php` (glob-based, no `pluginInfo.json` check), so extracting a shared helper is a reasonable idea for later, but there's no existing function this duplicates today.
- The help-page item: `www/help/plugins.php` (added to master since this branch was cut) documents the plugin-card privacy lights and buttons, but doesn't cover any header-level icon - and neither does the existing `navbarUpdateAvail` FPP-update icon this one is modeled after. That looks like a deliberate convention (global header chrome isn't page-specific F1 help), not a gap this PR introduces, so I left it alone rather than break the pattern.

Ready for another look.

darylc:
Verified the round-1 fixes — they hold up, and the regenerated openapi.json matches. Two changes needed before merge, then some small stuff.

**1. Switch the cache to `file_cache()` and drop the per-plugin bookkeeping.**

The current design keeps a per-plugin entry that five call sites (install, uninstall, upgrade ×2, updates-tab check) update under a lock. That's a lot of moving parts for a boolean, and two bugs in the revised code come straight from it:

- **Icon stays lit after a partial upgrade.** `UpgradePlugin()` only clears the entry on rc 0, but rc 2 means the code was pulled and only the post-script failed — the plugin is current, yet the icon stays lit for up to 6h. Both the streaming and JSON paths.
- **Writers block behind the poll's fetch.** `UpdatePluginUpdatesCacheEntry()` takes a blocking `LOCK_EX` while `GetPluginUpdatesAvailable()` holds the lock across ping + `git fetch` + `PluginHasUpdates()` (~20s), so an upgrade finishing or an Updates-tab check stalls behind a header poll. It can't just be narrowed either — the whole-array write at the end would overwrite anything written during the fetch.

Both disappear if the cache is one `{ts, updatesAvailable}` per box, built by `file_cache()` from `common.php` — it's already what `GetPluginList()` uses in this same file, and TTL, single-flighting and serve-stale-on-failure are all handled there — and the mutation sites just `@unlink()` the cache file so the next page load recomputes. Nothing to update out-of-band, nothing to lock against, nothing to get wrong on rc 2, and no third bespoke JSON-cache-plus-lockfile in this controller. Keep the ping gate, the `timeout 20`, and the one-fetch-per-call cap inside the builder — those are the right calls. I'd expect the PHP side to shrink to about a third of its current size.

**2. Use `InstalledPluginNames()`** for the installed-plugin list instead of the copy in `GetPluginUpdatesAvailable()`. It's in master since your merge, and its `sort()` also makes the one-per-poll rotation deterministic.

**Small, take or leave:**
- `cd X && $SUDO timeout 20 git fetch` (as `PluginFetchBranch()` does) rather than the `bash -c` wrapper.
- `menu.inc`: `foreach (glob(...))` warns on PHP 8 if `glob()` returns `false`; `(bool) glob(...)` is enough.
- A sentence in `www/help/plugins.php` under "Updates and reinstalls" saying what the header icon means and that FPP fetches each plugin's repo about every 6h.
- Call `checkForPluginUpdates()` from `UpdateAllFinish()` and `RunUpgradePlugin()`'s done callbacks so the icon clears on the page where the upgrade happened.

**To merge:** 1 and 2. Nothing under "Small" is blocking.

focusedonsound:
Thanks for the second pass — both blocking items done, plus the small stuff.

**1. Cache switched to `file_cache()`.** `GetPluginUpdatesAvailable()` is now one `{updatesAvailable}` flag for the whole box via `file_cache()` (the same helper `GetPluginList()` already uses in this file), not a per-plugin cache with its own lockfile. `UpgradePlugin()`, `UninstallPlugin()`, `InstallPluginFromInfo()`, and `CheckForPluginUpdates()` now just `@unlink()` the cache file on any change that could affect the answer — nothing to update out-of-band, nothing to lock against. This directly fixes both bugs you flagged:
   - **Partial-upgrade bug**: both `UpgradePlugin()` call sites now clear the cache on `$return_val != 1` (0 *or* 2 — 2 means the code landed and only the post-script failed, so the plugin genuinely isn't stale anymore), not just on 0.
   - **Writer-blocks-behind-poll bug**: gone by construction — mutation sites don't take any lock at all now.

   One addition beyond what was asked: since recompute now only happens once per 6h window box-wide (not once per page load), I kept the one-fetch-per-recompute cap but rotate *which* plugin gets it across successive TTL windows (`intdiv(time(), TTL) % count($plugins)`, no stored state) rather than always picking the same one — otherwise a box with several plugins would only ever live-refresh whichever one happens to sort first.

**2. `InstalledPluginNames()`** used instead of the duplicated directory scan.

**Small items:**
- `git fetch` now built `cd X && $SUDO timeout 20 git fetch` (matching `PluginFetchBranch()`'s style) instead of the `bash -c` wrapper.
- `menu.inc`: `(bool) glob(...)` instead of `foreach`-ing a result that can be `false`.
- `RunUpgradePlugin()` and `UpdateAllFinish()` now both re-poll the navbar icon immediately when they finish, instead of leaving it showing pre-upgrade state until the next full page load.
- Left the help-page item alone, per round 1's reasoning (global header chrome isn't page-specific F1 help, matching the existing `navbarUpdateAvail` icon this one is modeled after) — you didn't dispute that when you repeated it as "take or leave," so I'm reading that as settled rather than re-litigating it.

Also rebased onto current master (was 70-some commits behind again) and force-pushed nothing — this was a merge, and `git diff master...HEAD` is back to a clean 7-file, ~330-line diff.

Ready for another look.

darylc:
@focusedonsound As discussed, I've added this feature outside of your PR.  This turned out to be incredibly complex with a lot of moving parts to ensure consistency and safety.  Thanks for working on this.

PR #2946, opened 2026-09-12
feat(developer): Allow changing branch to an open Pull Request

jessica12ryan:
# feat(developer): Allow changing branch to an open Pull Request

## Summary
Adds **Pull Requests** as a new item inside the existing **Git Remote Repository** dropdown on `settings.php#settings-developer` (`www/settings-developer.php:256`, `www/settings.json:1715`). When selected, the existing **Git Branch** dropdown underneath is repurposed to list all open pull requests on `FalconChristmas/fpp` as `#<PRNUMBER> - <USER>` (truncated to ~60 chars, `[Draft]` prefix, drafts included). Selecting a PR checks out its branch via the same streaming upgrade path (`changebranch.php → scripts/git_branch`) as normal branch switches, using `origin pull/<n>/head → pr-<n>` so fork PRs work without adding per-fork remotes.

This was created with developers and testers in mind. When PR's are marked for review or testing, a developer or tester can now switch to the PR branch directly from the FPP UI. This makes it much easier for testing when a user creates a PR that requires others to physically test on different hardware. No git commands, no shell, no ssh required.

### `www/settings.json`
- Added `Pull Requests: pull-requests` option to `gitRemote` select and extended `tip` to describe PR mode.

### `www/settings-developer.php`
- `PrintGitBranchOptions()` early-returns for `pull-requests` (synthetic remote); renders `pr-<n>` if already on a PR branch, otherwise placeholder – no `git fetch -p --all` for PR mode.
- Added `reloadPullRequests()` → `GET api/git/pullRequests`, populates `#gitBranch` as `#<num> - <user>` (≤60 chars, `[Draft] ` prefix), `value=pr-<n>` + `data-pr`, `title` in tooltip.
- Added `formatPRLabel()`, `prCache`, `updatePRMeta()` – shows `PR # + title + user + headRef + View on GitHub` in `#prMeta` callout below the dropdown.
- Wrapped `window.ChangeGitBranch(branch, prNum)` to handle `pull-requests`: `confirm` then `location.href='changebranch.php?branch=pr-<n>&remote=pull-requests&pr=<n>'` (encoded); otherwise delegates to original `fpp.js:4239 ChangeGitBranch()`.
- `$('#gitRemote').on('change')` now dispatches to `reloadPullRequests()` for `pull-requests`, else `reloadGitBranches()` (origin/newfeatures/fork).
- Initial load: if `$('#gitRemote').val() === 'pull-requests'` calls `reloadPullRequests()` immediately. Added `#prMeta` container.

### `www/api/controllers/git.php`
- Added `GitPullRequests()` – `GET /api/git/pullRequests` proxies `https://api.github.com/repos/FalconChristmas/fpp/pulls?state=open&per_page=100&sort=updated` with `User-Agent: FPP`, optional `Authorization: token <gitHubPAT>` (CRLF-checked). Returns `[{number,title,user,headRef,headRepo,headSha,htmlUrl,draft}]`, empty array on non-2xx/offline. Handles `head.repo == null` (deleted fork).
- `GitBranches()` now maps synthetic `pull-requests` → `origin` before filtering, so callers without explicit `?remote=` don't get empty results.

### `www/api/index.php`
- Added `dispatch_get('/git/pullRequests','GitPullRequests')`.

### `www/api/openapi.json`
- Added `/api/git/pullRequests` spec.

### `www/changebranch.php`
- Accepts `remote=pull-requests` as valid synthetic remote (bypasses `/^[a-zA-Z0-9_-]+$/` regex).
- Extracts `prNumber` from `?pr=` or `pull/<n>/head` or `pr-<n>`, validates `>0`.
- Locked `git fetch origin pull/<n>/head:refs/remotes/origin/pull/<n>/head` (via `flock -w 60 /tmp/fpp-git-repo.lock` consistent with existing fork handling), then normalizes to `branch=pr-<n>, remote=origin` before invoking `scripts/git_branch`.
- Fork handling (`origin`/`newfeatures` excluded) now also excludes `pull-requests` via `!$prFetchDone`.

### `scripts/git_branch`
- Detects `^pr-[0-9]+$` as PR local branch. Validates `refs/remotes/<remote>/pull/<n>/head` via `git show-ref -F`; on success sets `PR_LOCAL`.
- New `if [ -n "$PR_LOCAL" ]` path: `git clean -df && reset --hard && submodule reset && git fetch <remote> pull/<n>/head:refs/remotes/<remote>/pull/<n>/head | head -20 && git checkout -B pr-<n> <remote>/pull/<n>/head && submodule sync/update`. Normal `origin`/`newfeatures`/fork branches keep existing `rebase` flow unchanged.

### `www/help/settings-developer.php`
- Updated **Git Remote Repository** and **Git Branch** bullets to document `Pull Requests` → `pull-requests` via `api/git/pullRequests`, `#<num> - <user>` ≤60 chars + `[Draft]`, `pr-<n>` checkout via `origin pull/<n>/head`, `git clean -df` warning.

## Behavior
- Consistency: same confirm → `changebranch.php` streaming → `git_branch` → `upgrade_config` → rebuild path as normal branch switches.
- PRs work for both intra-repo and fork PRs: uses `origin`'s `pull/<n>/head` ref, no per-fork remote pollution.
- Empty/offline/rate-limited GitHub → dropdown shows `No open pull requests` / `Failed to load PRs` with info callout, not a hard error.
- Selecting a PR shows meta panel with PR title, draft badge, author, head ref, and GitHub link.
- Existing `origin`/`newfeatures`/fork remotes and branches unaffected; `gatherStats`/default unchanged.

## Testing
- `php -l` on `www/api/controllers/git.php`, `www/changebranch.php`, `www/settings-developer.php`; `bash -n scripts/git_branch`; `python -m json.tool` on `www/settings.json` and `www/api/openapi.json` – all pass.
- Manually and throughly tested on Pi 4B

## Risk / Backwards Compatibility
- Additive: new `pull-requests` option; `GitBranches` fallback to `origin` for that value prevents empty listings for callers without explicit `?remote=`.
- No public header/plugin interface changes.
- Fork handling unchanged; multisync's `branchRemoteSelect` (origin/newfeatures only) unaffected when stored setting is `pull-requests`.

## Files Changed
- `www/settings.json`
- `www/api/controllers/git.php`
- `www/api/index.php`
- `www/api/openapi.json`
- `www/changebranch.php`
- `scripts/git_branch`
- `www/settings-developer.php`
- `www/help/settings-developer.php`

## Screenshot
ss_featDevPR


OnlineDynamic:
Thanks for this — the concept is genuinely useful, and being able to hand a tester a PR number instead of ssh instructions solves a real problem. I went through it carefully, including the git ref semantics, and verified the behaviours below in a scratch repo rather than assuming.

The patch applies cleanly and `bash -n` / `php -l` / JSON parse all pass on the eight changed files. The help page and the `settings.json` tip were both updated, which is appreciated.

Two things I specifically checked and want to confirm are **fine**:

- **Remote URLs are never touched.** There's no `git remote add`/`set-url` on the PR path — it fetches by explicit refspec from the existing `origin`, and `changebranch.php` rewrites `$remote` back to `origin` before the fork-remote block (with the extra `!$prFetchDone` guard). `pull-requests` can't reach `scripts/git_branch` as a REMOTE either, since `$prNumber <= 0` 400s out first.
- **Recovery works.** Switching the remote back to FPP Releases and picking a branch runs the normal `checkout -B` / `rebase` path cleanly. Nobody gets stranded on a PR branch.

There's one issue I'd consider blocking, plus some smaller items.

---

## Blocking: force-push leaves the device on a stale commit, and reports success

The refspec has no `+`, so it's fast-forward-only. PR branches get force-pushed constantly (rebases, amends, squashes), and every one of those makes the fetch non-fast-forward. Reproduced:

```
 ! [rejected]  refs/pull/1/head -> origin/pull/1/head  (non-fast-forward)   rc=1
local ref still: 34b98a5     (PR head is actually dc54db4)
```

`changebranch.php` prints "Failed to fetch PR #N" and then deliberately continues; `git_branch`'s `show-ref` check *passes* on the now-stale ref, checks it out, rebuilds, and the whole switch reports success. The tester ends up building an old revision of the PR and reporting results against code that no longer exists — which is the exact failure this feature is meant to prevent. Re-selecting the same PR doesn't recover, because the fetch is rejected identically every time.

Fix is one character in each of two places:

**`scripts/git_branch`**
```diff
-git fetch $REMOTE pull/${PR_NUM}/head:refs/remotes/${REMOTE}/pull/${PR_NUM}/head
+git fetch $REMOTE +pull/${PR_NUM}/head:refs/remotes/${REMOTE}/pull/${PR_NUM}/head
```
(in both the `/opt/fpp` and the `sudo -u ${FPPUSER}` variants)

**`www/changebranch.php`**
```diff
-$fetchInner = "git --git-dir=$gitDir fetch origin pull/" . $prNumber . "/head:refs/remotes/origin/pull/" . $prNumber . "/head 2>&1";
+$fetchInner = "git --git-dir=$gitDir fetch origin +pull/" . $prNumber . "/head:refs/remotes/origin/pull/" . $prNumber . "/head 2>&1";
```

Since the whole switch is preceded by `git clean -df && git reset --hard` anyway, forcing the ref update is consistent with what the branch path already does.

---

## Worth fixing: "Update FPP" silently does nothing on a PR branch

`git checkout -B pr-N origin/pull/N/head` sets `branch.pr-N.merge = refs/heads/pull/N/head`, a ref that doesn't exist on origin. A plain `git fetch` uses `+refs/heads/*:refs/remotes/origin/*`, which never refreshes `refs/remotes/origin/pull/N/head`. So `git_pull`'s `git fetch && git rebase @{u}` rebases onto a frozen local ref forever — it recompiles and reports success without ever pulling anything new.

Not destructive, but confusing for a tester who hits Update expecting the latest PR head. Teaching `git_pull` to re-fetch when HEAD looks like a PR branch would make the feature coherent:

```bash
if [[ "${GITBRANCH}" =~ ^pr-([0-9]+)$ ]]; then
    PR_NUM="${BASH_REMATCH[1]}"
    runGitLocked "cd ${FPPDIR} && git fetch origin +pull/${PR_NUM}/head:refs/remotes/origin/pull/${PR_NUM}/head"
fi
```

Related and fails safe, so just flagging it: `get_remote_git_version()` in `www/common.php` ends up "Unknown" on a PR branch, because `git ls-remote -q -h origin pull/N/head` returns nothing (`-h` is heads-only) and the `pr-N` fallback misses too. No bogus "update available" prompt, but the status page looks broken while you're on a PR.

---

## Smaller items

- **`scripts/git_branch` validation ordering** — the PR `show-ref` check runs *before* the script's own fetch, so `scripts/git_branch pr-123 origin` from a shell always fails with "fetch PR first". It only works today because `changebranch.php` pre-fetches. Moving the check after the fetch (or dropping it, since the fetch is authoritative) makes the script usable standalone.
- **Missing `git fetch --tags -f`** — the normal branch path has it, the PR path doesn't. `www/common.php:2696` uses `git describe`, so the head-version display goes stale after switching to a PR.
- **Inconsistent escaping in `settings-developer.php`** — `formatPRLabel()` interpolates `pr.user` unescaped, and `updatePRMeta()` puts `pr.htmlUrl` straight into `href`, while `pr.title` right next to them *is* escaped via `$('<div>').text(...).html()`. Both values come from GitHub so it isn't exploitable, but worth making consistent.
- **`GitPullRequests()` masks errors and isn't cached** — every failure returns `[]` with HTTP 200, so an expired `gitHubPAT` (401) is indistinguishable from "no open PRs", which will generate support questions. Two suggestions: retry unauthenticated on 401/403, and wrap the call in `file_cache` (~60s). Unauthenticated GitHub is 60 requests/hour *per public IP*, shared across every FPP behind one NAT, and the endpoint currently fires on every page load in PR mode.
- **`window.ChangeGitBranch` monkey-patch** — works, but it's fragile to depend on load order. A `remote === 'pull-requests'` branch inside `ChangeGitBranch` in `www/js/fpp.js` would be more direct.
- **No cleanup** — local `pr-*` branches and `refs/remotes/origin/pull/*/head` refs accumulate indefinitely with no pruning.
- Minor: `www/multisync.php:4123` reads `gitRemote` to preselect its remote dropdown, and `pull-requests` matches neither option so it falls back to showing origin. Harmless, just noting it.

---

Happy to re-review once the refspec fix is in — that one is the difference between "the tester is on the PR" and "the tester is silently on a stale revision of it." Everything else can reasonably follow in a later pass.

jessica12ryan:
### Applied review fixes (@OnlineDynamic) – production-ready, non-breaking, verified php -l/bash -n/JSON

#### Blocking – force-push stale ref:

 - ```www/changebranch.php:62``` +pull/<n>/head:refs/remotes/origin/pull/<n>/head
 - ```scripts/git_branch:69,71``` both variants +pull/${PR_NUM}/head:refs/remotes/${REMOTE}/pull/${PR_NUM}/head (also added git fetch --tags -f missing in PR path). Re-selecting same PR now force-updates stale ref instead of non-fast-forward rejected → success on stale commit.

#### PR branch update (git_pull):

 - ```scripts/git_pull:43``` after GITBRANCH detection, if ^pr-([0-9]+)$ runs runGitLocked git fetch origin +pull/<n>/head:refs/remotes/origin/pull/<n>/head (both ```/opt/fpp``` and sudo -u) so git fetch && git rebase @{u} rebases onto fresh PR head instead of frozen ref.
 
#### Version display (```www/common.php:2783```):

 - get_remote_git_version() retry without -h for pr-*/pull/<n>/head tracking (ls-remote -q $remote pull/<n>/head via escapeshellarg), fixes “Unknown” on PR branch (-h is heads-only).

#### Validation ordering (```scripts/git_branch:40```):

 - Removed pre-fetch show-ref check for pr-*; set PR_LOCAL on name match and validate via checkout after fetch, so git_branch pr-123 origin works standalone (previously always “fetch PR first” unless ```changebranch.php``` pre-fetched).

#### Escaping consistency (```www/settings-developer.php:167,202```):

 - formatPRLabel() now escapes via $('<div>').text(raw).html() (≤60 chars), updatePRMeta() validates htmlUrl regex ^https://github.com/FalconChristmas/fpp/pull/\d+ and escapes title/user/headRef/htmlUrl consistently.
 
#### GitPullRequests() caching & error masking (```www/api/controllers/git.php:646```):

 - Wrapped in file_cache('git_pull_requests',…,60,10) (returns null on failure to preserve last-good, vs poisoning cache with []). Retry unauthenticated on 401/403 if PAT was sent (expired PAT no longer masquerades as “no PRs”).

#### ChangeGitBranch fragility (```www/js/fpp.js:4239```):

 - Added direct remote==='pull-requests' branch handling (parses data-pr/pr-*/pull/<n>/head, confirm for PR, encodeURIComponent for both params) – primary path no longer depends on monkey-patch load order. ```settings-developer.php``` wrapper kept as compatible fallback.

### Verification
 
Verified no breaking: gitRemote=pull-requests still synthetic, GitBranches() fallbacks pull-requests→origin, multisync remote dropdown unchanged (harmless fallback), fork paths unchanged, no header/plugin interface changes.

### Additional Comments

I have not had the time to test this with the fixes added. I've got a busy day today and I know I wouldn't be available to do this for quite a while so I just wanted to get the fixes thrown in so it could atleast be reviewed.

PR #2949, opened 2026-09-12
BBShiftString: keep a V4 (send-only) chain's port 1 off during the packet phase

### LINKED ISSUE #2948: BBShiftString: a FalconV5 chain's query frames corrupt a FalconV4 chain on the same cape (packet phase is cape-wide)
**Setup.** K32-Max, FPP 10.0 with `826802ab8` applied (the #2854 fix) plus #2881. Ports 33-36: FalconV4 smart receiver (a Genius 4-output long-range receiver, 2 strings of ~700 px). Ports 37-40: FalconV5 SRx1 (2 strings, 400 + 585 px). The master is a separate FPP player feeding the K32 as a remote over multisync at 40 fps.

**Symptom.** V4 chain alone: fine. V5 chain alone: fine. Both configured: the V5 chain works, **the V4 chain shows garbage, worst on the tail pixels of its strings.** Unusable for a show.

**Cause, from the code.** `FalconV5Support::generateDynamicPacket()` schedules a query (and a listen window) for the V5 chain every other frame. `BBShiftStringOutput::encodeFalconV5Packet()` fills the packet slot for *every* string of the cape, and the PRU macro `OUTPUT_FALCONV5_PACKET` then clocks the preamble (15.3 us low, 53.5 us high) and the 57 bytes out on all pins at once; there is no per-pin gate in the packet phase. V4 chains are kept out of the query schedule (`sendOnly`, from `826802ab8`), so their slot is the zero-filled buffer. A V4 receiver therefore sees, every other frame, a framed packet of 57 zero bytes that it takes as a config packet, and a preamble it takes as a chain break. Alone, a V4 chain meets the packet phase only for its own config packets at startup, which is why it works alone.

**What was tested on the box.**

1. Give every send-only chain a valid config packet in every query frame instead of zeros: slightly calmer, still garbage on the tail pixels. The packet phase itself is the problem, not only its content.
2. Treat V5 chains as send-only too (no query frames at all): both chains clean, V5 still passing data. Proves the trigger, but costs all V5 telemetry.
3. **The fix below:** V5 chains stock, the V4 chain's port 1 switched off for the packet phase, exactly as ports 2-4 of every chain already are. Both chains clean, V5 chain queried and reporting currents.

**Fix.** In `BBShiftStringOutput::setupFalconV5Support()`, for `sendOnly` chains add a GPIO command that turns port 1 off at the end of the pixel data:

```cpp
            p1->m_gpioCommands.clear();
            if (sendOnly) {
                // A V4 chain is never queried, but the packet phase is clocked out
                // on every pin of the cape. Keep its port 1 off during that phase,
                // as ports 2-4 already are, or a V4 receiver next to a V5 chain
                // reads the V5 query packets as a chain break plus pixel data.
                p1->m_gpioCommands.emplace_back(1, max, 0, 0);
            }
```

**Verification.** Five sequences at 40 fps from the master over multisync, three of them with the real props (1,392 px on the V4 chain over two outputs, 985 px on the V5 chain over two ports). Frame rate from the K32's own per-second `FrameNum` log line:

| Run | Load | fps mean | lowest second | gate declined | V5 peak current | errors |
|---|---|---|---|---|---|---|
| Ghostbusters, test strings | 130 px per chain | 39.99 | 39.72 | 0 / 6,000 | 1,363 mA | 0 |
| Ghostbusters, test strings | 130 px per chain | 39.99 | 39.80 | 0 / 6,000 | 1,287 mA | 0 |
| Ghostbusters, real props | 1,392 + 985 px | 39.99 | 39.80 | 0 / 7,541 | 2,895 mA | 0 |
| Stranger Things Remix | 1,392 + 985 px | 39.98 | 39.80 | 0 / 6,431 | 2,262 mA | 0 |
| Alien Ant (594 effects on the V4 prop) | 1,392 + 985 px | 39.98 | 39.80 | 0 / 6,000 | 878 mA | 0 |

Both chains visually clean in every run; before the fix the V4 chain was visibly corrupted within seconds in test mode.

**Open point.** With the mask the V4 chain also never receives its startup config packet. The Genius receiver does not need one, tested. If a V4 receiver exists that does, the mask would have to be lifted for the initial config packets only (a two-stage command table), or the config packet moved ahead of the mask point. I do not have such a receiver to test.

**Files.** `src/non-gpl/BBShiftString/BBShiftString.cpp` (`setupFalconV5Support`, `encodeFalconV5Packet`), `src/non-gpl/BBShiftString/BBShiftString.asm` (`OUTPUT_FALCONV5_PACKET`), `src/non-gpl/FalconV5Support/FalconV5Support.cpp` (`generateDynamicPacket`, `addReceiverChain`). Follow-up to #2854 / `826802ab8`. PR follows.

Starfield3:
Fixes #2948.

A FalconV5 chain's query frames are clocked out on every pin of the cape; a FalconV4 chain on another group reads them as a chain break plus pixel data and shows garbage on its tail pixels. Ports 2-4 of every chain are already switched off for the packet phase through the GPIO command table; this does the same for port 1 of a send-only chain, which is never queried and has nothing to receive in that phase.

Seven lines in `setupFalconV5Support()`, no PRU change, no new setting.

Verified on a K32-Max with a Genius V4 receiver (1,392 px over two outputs) and an SRx1 v5 (985 px over two ports) on the same cape: five 40 fps sequences from a master over multisync, 39.98-39.99 fps mean, no second below 39.8, zero frames declined by the back-pressure gate, V5 telemetry live throughout, V4 chain clean. Details and the two experiments that narrowed it down are in #2948.

Open point, stated in the issue: the V4 chain then never receives its startup config packet. The Genius does not need one; a V4 receiver that does would need the mask lifted for the initial config packets.

### DIFF
--- a/src/non-gpl/BBShiftString/BBShiftString.cpp
+++ b/src/non-gpl/BBShiftString/BBShiftString.cpp
@@ -1813,6 +1813,13 @@ void BBShiftStringOutput::setupFalconV5Support(const Json::Value& root, uint8_t*
             // and must have edges that are aligned.  Need to turn OFF the 2-4 ports during
             // the config packet
             p1->m_gpioCommands.clear();
+            if (sendOnly) {
+                // A V4 chain is never queried, but the packet phase is clocked out
+                // on every pin of the cape. Keep its port 1 off during that phase,
+                // as ports 2-4 already are, or a V4 receiver next to a V5 chain
+                // reads the V5 query packets as a chain break plus pixel data.
+                p1->m_gpioCommands.emplace_back(1, max, 0, 0);
+            }
             if (p2) {
                 p2->m_gpioCommands.clear();
                 p2->m_gpioCommands.emplace_back(2, max, 0, 0);


dkulp:
Replaced with a proper fix.

PR #2960, opened 2026-09-14
BBShiftString: FalconV5QueryInterval — space receiver queries and read the reply a frame later (experimental, default off)

### LINKED ISSUE #2959: BBShiftString: FalconV5 receivers at 768 px/port @ 40 fps — stock 10.0 storms (phantom eFuse trips, frame skips); query spacing + a one-frame reply delay makes it run clean (K32-Max)
**In one line.** Long strings behind Falcon V5 smart receivers do not work on FPP 10.0 at 40 fps: at ~740+ px per port the receiver's reply is read one frame too early, decodes as garbage, and every query turns into phantom `eFUSE Triggered` events and frame skips. Spacing the queries and reading the reply one frame later makes 768 px/port run at a flat 40 fps with telemetry alive — measured below, including on unmodified 10.0.

**Setup.** K32-Max (AM62x), FPP v10.0, 40 fps content from a master over multisync, garage matrix 6 × 768 + 192 px behind smart receivers on cape ports 33–39 — on 13 Sep one SRx1 v5 (33–36) beside a Genius V4 (37–39, with the PR #2949 mask), on 14 Sep two SRx1 v5. Same rig, same receiver, 400 + 585 px per port on 12 Sep: clean.

**Symptom.** Frame rate 2.8–67 fps per second, `Send:` up to 554 ms, "Skipping many frames", CPU 3 % idle. The log shows `eFUSE Triggered for Port 33A/34A/35A/36A` — all four within ~40 ms, 96 events in one 3-minute song, 649 that day — each self-clearing 0.12–0.8 s later. A port snapshot 10–100 ms after each event: all four `status: true`, `enabled: true`, 170–1,900 mA. The trips are not real.

**Mechanism.** (1) `FalconV5Support::generateDynamicPacket` schedules a query on every other free config slot. A query frame costs the cape the packet phase (57 bytes at 1,130 ns/bit ≈ 0.75 ms) plus the response window (`OUTPUT_FALCONV5_PACKET`: `SLEEPNS 500000` ×4 + 100000 ≈ 2.15 ms when data is found) on every pin. At 768 px (23.0 ms of data) the next 25 ms tick lands ~1.2 ms inside that window. (2) `BBShiftStringOutput::SendData` calls `processListenerData()` first thing, which reads `pruData->length` and the capture buffer while the listener is still storing the reply, then zeroes `length`. The partial packet passes `decodeFalconV5Packet` and sets fuse flags for all four ports of that receiver → four warnings, four `EFUSE_TRIGGERED` preset runs, each a `fork` inside the output thread. (3) Separately, the ~2.9 ms per query frame exceeds the ~0.8 ms per-frame slack at 768 px, so the back-pressure gate declines 2.3 % of frames.

**What we changed (experiment, 46 lines — PR to follow):** a setting `FalconV5QueryInterval` (frames between query packets, 1 = current behaviour; sticky "owed" flag in `PrepData`, config/number packets untouched), and with interval > 1, `SendData` skips `processListenerData()` on the frame right after the one that carried the query — the reply is read complete one frame later.

**Reproduced on unmodified 10.0 (14 Sep, tag 370e62e, both libs rebuilt from the tag):** two SRx1 v5 (cape 33–36 and 37–40), 6 × 768 + 192 px, 40 fps playlist: within 20 s the log fills with `eFUSE Triggered` — 158 events in 99 s (33A–36A 9 each, 37A–40A 30–32 each, **including port 40A, which has 0 pixels**), 55 "Skipping many frames", `Send:` up to 551 ms, 26–55 fps. Same run with the patched build and the setting at 1 (= stock path): 201 events / 66 skips in 100 s. Same run with the setting at 40: **31:45 clean** — 0 eFUSE, 0 skips, 30 of 79,818 frames declined (29 at the cold start, 1 at a file open), 39.97 fps mean, all six receiver ports reporting throughout (peaks up to 4.3 A).

**Result, same rig, interval 40 (13 Sep):** Ghostbusters + Stranger Things back to back, 416 s: 39.98 fps mean, min 39.80, gate 0 of 11,998, eFUSE 0, skips 0, receiver telemetry alive (167–188 value changes per port), CPU idle ~90 %. Repeated with the V4 receiver loaded (425 s): identical. Stock cadence (interval 1) reproduces the storm within a minute.

**Proposal.** Ship the interval as an *Experimental*-level setting (default 1, nothing changes for anyone) together with the delayed read, or derive it from the frame-budget model (#2886: `m_frameTimeUs` vs. period → interval 1 when the query fits, larger otherwise). Open question for you: would you rather the listener PRU flag "capture complete" than a frame delay? Side note: presets triggered from `OutputMonitor::triggerEFusePresets()` run synchronously in the output thread — a user's script (ours: four `curl` per burst) turned four fake trips into a half-second output stall. Worth running them off-thread regardless of this bug.

Logs, snapshots and the per-frame measurements are available on request.

Related: #2948 / #2949 (V4 chain corrupted by V5 query frames on the same cape — the other half of running long strings on mixed receivers), #2886 (frame budget model).

Starfield3:
Fixes #2959 — or rather, makes the case there runnable behind a switch that is off by default.

**What it does.** New setting `FalconV5QueryInterval` (Experimental level, default 1 = today's behaviour): the number of frames between dynamic (query / pixel-count / eFuse) packets to Falcon V5 receivers on BeagleBone capes. With an interval > 1:

- a query is *owed* every N frames and sent on the next free config slot (sticky flag — a config re-send or a mux-change wait delays it rather than drops it); config and number packets keep their slots;
- `processListenerData()` is skipped on the frame right after the one that carried the query, so the receiver's reply is read one frame later, complete, instead of mid-capture.

At interval 1 every added branch takes the stock path — verified by running the patched build with the setting at 1 (storm reproduced identically to unmodified 10.0).

**Why.** At ~740+ px per port and 40 fps the next frame tick lands inside the receiver's response window; the partial reply decodes into fuse flags for all four ports of that receiver, and every query becomes a burst of phantom `eFUSE Triggered` warnings (each running the `EFUSE_TRIGGERED` preset inside the output thread) plus frame skips. Details and measurements in #2959.

**Measured** (K32-Max, FPP 10.0, two SRx1 v5, 6 × 768 + 192 px, 40 fps playlist): stock — 158 phantom trips and 55 frame skips within 99 s, 26–55 fps; interval 40 — 31:45 clean, 0 trips, 0 skips, 30 of 79,818 frames declined (all at the cold start and one file open), 39.97 fps mean, all six receiver ports reporting throughout. Cost at 40: port current readings update ~once a second instead of ~20×.

**Notes for review.** The edits are the same six anchored changes that ran on 10.0; master's code at those spots is identical, so this is a rebase, not a port — but I could not build master here (the box runs 10.0), so please treat it as "compiles on 10.0, logic mirrors the tested patch". Two alternatives you may prefer, both fine by me: derive the interval from the #2886 frame-budget model instead of a setting (interval 1 whenever the query fits, larger otherwise), or have the listener PRU flag "capture complete" instead of a fixed one-frame delay. The `EFUSE_TRIGGERED` presets running synchronously in the output thread is a separate thing worth its own look.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dkulp:
This is no longer necessary (and wouldn't have actually solved it, just made it 40x less likely to happen)

